PR #1, opened 2016-04-12
[CUBRIDQA-5] Use the 'Asia/Shanghai' instead of 'Africa/El_Aaiun', to make result stable.

jinmeihua:


slamdunkorchid:
no problem,please merge it.

PR #2, opened 2016-04-14
[CUBRIDQA-4] add exclude list for ha_repl test

slamdunkorchid:
1. add sql/conf/daily_regression_test_ha_repl.conf 
since there are 7290 cases, I only make 62 cases skip according regression test


fanzaiqiang:
Dear Orchid,

Please commit it.

Also commit to release/10.0 branch.

PR #5, opened 2016-04-27
[CUBRIDQA-4] add exclude list for compat test

slamdunkorchid:
Dear Fan,
1.I have added exclude list and patch files for sql case
2.mv some file to medium


fanzaiqiang:
Dear Orchid,

Please commit it. Thank you.

PR #21, opened 2016-06-03
[CUBRIDQA-4] modify patch files,since the difference result of jdk1.6 and jdk1.7

slamdunkorchid:
Dear May,
Please help me to review it.

reason:
table column euckr used together with jdbc(:::?charset=utf-8)
jdk1.6 and jdk1.7 have different result.
This also appears on 9.3server—>9.3driver
It is the problem for all legacy version, It has changed on 10.1
And euc-kr’s usage is not recommended

the test environment use 
jdk1.6.0_22


jinmeihua:
Okay, please merge it.

PR #23, opened 2016-06-27
[CBRD-20031] add cases for the fix of bad results using Like

slamdunkorchid:
Dear Colin,
Please help me to review it.


supler:
@slamdunkorchid 

how about to extent the testing point? for instance as the below to confirm the values

prepare st from 'select * from t where c like ?' 
execute st using '%';

supler:
@slamdunkorchid 

please commit it,thanks

PR #27, opened 2016-07-11
[CBRD-20140][CBRD-20085] add test case for trim() fucntion with binary and euckr charset.

jinmeihua:


slamdunkorchid:
please commit it.

PR #29, opened 2016-07-25
[CBRD-20031][CBRD-20085] banana i18n spec changed,add binary,so do patch for leaacy

slamdunkorchid:


jinmeihua:
@slamdunkorchid 
OK, please merge it.

PR #35, opened 2016-09-08
[CBRD-20465]Add test case for 'DDL Query with C-Style comments do not…

jinmeihua:
… replicate'.


slamdunkorchid:
Please go ahead!

PR #39, opened 2016-10-05
[CUBRIDQA-69] TC update for timezone

kisoo-han:
In the test case _02_addmonths.sql of
sql/_27_banana_qa/issue_5765_timezone_support/_08_current_date_function/_02_functions

Case should be modified to resolve timezone difference.


xkyu:
Please do not mix two or more issues on a PR.

PR #44, opened 2016-10-17
[CBRD-20399] Extension of DEFAULT Clause in CREATE statements

kisoo-han:
This is a set of TestCase for [CBRD-20399] Extension of Default Clause in CREATE statements.
More TestCases will be added with Code Coverage Analysis.


slamdunkorchid:
there are some changes based on latest build. 
fixed it as #256

PR #47, opened 2016-11-08
[CBRD-20572] add test cases for CBRD-20572

mandy2016:


jiyanping:
@mandy2016 
Please add 'drop table t1,t2,t3;' in the end of the test case.

jiyanping:
@mandy2016 
Please merge it,thank you.

PR #48, opened 2016-11-09
[CBRD-20576] add test case for CBRD-20576

jiyanping:
http://jira.cubrid.org/browse/CBRD-20576


mandy2016:
@jiyanping 
please commit it.

PR #54, opened 2016-12-02
[CUBRIDQA-12]relate to current_date,make it stable

slamdunkorchid:


supler:
@slamdunkorchid 

please go ahead, thanks

PR #64, opened 2017-01-09
[CBRD-20595] Add test case for CBRD-20595

jiyanping:


jiyanping:
http://jira.cubrid.org/browse/CBRD-20595

mandy2016:
@jiyanping 
Please commit it.

PR #66, opened 2017-01-10
[CBRD-20596] Add test case for CBRD-20596

jiyanping:
http://jira.cubrid.org/browse/CBRD-20596


mandy2016:
@jiyanping 
please commit it

PR #68, opened 2017-01-11
[CBRD-20594] Add test case for CBRD-20594

jiyanping:
http://jira.cubrid.org/browse/CBRD-20594


mandy2016:
@jiyanping 
please commit it.

PR #69, opened 2017-01-12
[CBRD-20613] Add test case for CBRD-20613

jiyanping:
http://jira.cubrid.org/browse/CBRD-20613


mandy2016:
@jiyanping 
please commit it.

PR #70, opened 2017-01-13
[CBRD-20583] Add test case for CBRD-20583

jiyanping:
http://jira.cubrid.org/browse/CBRD-20583


mandy2016:
@jiyanping 
please commit it.

PR #77, opened 2017-02-09
[CBRD-20909] Add test cases to test memory leak of select_btree_node_info scan

jiyanping:
Add test cases to test memory leak of select_btree_node_info scan.
http://jira.cubrid.org/browse/CBRD-20909


jinmeihua:
@jiyanping ,
It would be better to add 'order by' clause to the select queries.

jiyanping:
@jinmeihua 
I tried to add 'order by' clause,but core dumped. I opened a new issue,
http://jira.cubrid.org/browse/CBRD-20923

PR #79, opened 2017-02-09
[CBRD-20820] add test cases for the fix 'pt_resolve_cte_specs: do not stop parse tree walk'

slamdunkorchid:
refer to http://jira.cubrid.org/browse/CBRD-20820
Test Points:
selecting rows using  IN/EXISTS and subquery of  CTE type


supler:
@slamdunkorchid 

please go ahead, thanks

PR #84, opened 2017-02-10
[CBRD-20824] add cases for the fix: Avoid dummy select optimization when SELECT includes WITH clause

slamdunkorchid:
refer to http://jira.cubrid.org/browse/CBRD-20824
Test Points:
selecting rows from table which generated by subquery of CTE type


supler:
@slamdunkorchid 

please go ahead, thanks

PR #85, opened 2017-02-10
[CBRD-20828] add cases for the fix:recursive list id leak

slamdunkorchid:
refer to http://jira.cubrid.org/browse/CBRD-20828
Test Points:
after appears error 'Overflow occurred in multiplication context.',not server crash


supler:
@slamdunkorchid 

please go ahead,thanks

PR #86, opened 2017-02-10
[CBRD-20830] add cases for the fix:Removed PT_LIST_WALK from pt_resolve_spec_to_cte

slamdunkorchid:
http://jira.cubrid.org/browse/CBRD-20830

Test Points:
reference the same CTE table multiple times,eg:
columns generated by subquery of CTE type
FROM table generated by subquery of CTE type
WHERE condition using subquery of CTE type


supler:
@slamdunkorchid 

please go ahead, thanks

PR #87, opened 2017-02-10
[CBRD-20835] add cases for the fix:Extend condition to differentiate entity specs from derived table/cte

slamdunkorchid:
http://jira.cubrid.org/browse/CBRD-20835

CTE spec attributes were resolved as for an entity spec. Extending condition should handle CTE specs correctly.


supler:
@slamdunkorchid 

please go ahead, thanks

PR #89, opened 2017-02-13
[CBRD-20843] add cases for the fix: Consider correlation level of CTEs when generating XASL aptr list

slamdunkorchid:
http://jira.cubrid.org/browse/CBRD-20843

excute below statements:
SELECT * FROM (WITH a AS (SELECT * FROM t1) SELECT 1) AS t1;


supler:
@slamdunkorchid 

please go ahead, thanks

PR #91, opened 2017-02-13
[CBRD-20838]  add cases for the fix:Bind CTE names and types separately, prior to the query binding

slamdunkorchid:
http://jira.cubrid.org/browse/CBRD-20838
CTE column definition correct,such as the number of column

create table t1 (a int, b  varchar(32));
create table t2 (c int);
with t(f) as (select * from t1 where b >= 'c') 
  select t1.b from t2,t1 where t1.a = t2.c;


supler:
@slamdunkorchid 

please go ahead,thanks

PR #92, opened 2017-02-13
[CBRD-20916] add cases for the fix:use pt_resolve_cte_specs as post function

slamdunkorchid:
http://jira.cubrid.org/browse/CBRD-20916

test below statements:
with  t as (select c from t2 where c >= 4)
  select t1.a,t1.b from t1,t
    where t1.a=t.c and
          t.c in (with  t as (select * from t1 where t1.a<5)
                    select t2.c from t2,t where t2.c=t.a);


supler:
@slamdunkorchid 

please go ahead, thanks

PR #93, opened 2017-02-13
[CBRD-20881] add cases for the fix:Bind CTE names and types separately

slamdunkorchid:
http://jira.cubrid.org/browse/CBRD-20881


supler:
@slamdunkorchid 

please go ahead, thanks

PR #97, opened 2017-02-14
[CBRD-20823] add case to examine ambiguous error when subquery of cte has same column name

supler:
http://jira.cubrid.org/browse/CBRD-20823

add case to examine ambiguous error when subquery of cte has same column name


fanzaiqiang:
@supler 

Please go ahead to merge. Thank you.

PR #103, opened 2017-02-15
[CBRD-20447] Add test case to test disk_size function

jiyanping:
Add test case to test disk_size function.
http://jira.cubrid.org/browse/CBRD-20447


sql/_30_banana_pie_qa/issue_20447_disk_size/cases/20447_01.sql
```diff
@@ -0,0 +1,78 @@
...
+
+
+-- session variable
+set @a=repeat('a',300);
+
+select @a,disk_size(@a);
+
+select @a,disk_size(cast(@a as char(100)));
+
+
```
mandy2016: 1. add more string functions in disk_size(), such as disk_size(substr()), disk_size(rpad()), disk_size(rtrim())
2. add some cases that using index, such as 
create index on t1(c1);
insert into t1 values(repeat('abc',100));
select disk_size(c1) from t1 where c1 like 'abc%'
3. add some cases to test more data types, such as float, decimal, set, enum

jiyanping: About the enum,set data types,it will be tested when I verify string compression.
http://jira.cubrid.org/browse/CBRD-20158

mandy2016:
@jiyanping 
please merge it.

PR #111, opened 2017-02-16
[CBRD-20955] add cases for the fix:Isolate current CTE from CTE list when resolving self references

slamdunkorchid:
http://jira.cubrid.org/browse/CBRD-20955


supler:
@slamdunkorchid 

please go ahead, thanks

PR #113, opened 2017-02-16
[CBRD-20941] add cases for the fix:Avoid recursive CTE optimization for rownum

slamdunkorchid:
http://jira.cubrid.org/browse/CBRD-20941


supler:
@slamdunkorchid 

please add order by

supler:
@slamdunkorchid 

please go ahead, thanks

PR #115, opened 2017-02-16
[CBRD-20935] add cases for the fix:Allow only PT_SELECT node as the recursive query of CTEs

slamdunkorchid:
http://jira.cubrid.org/browse/CBRD-20935


supler:
@slamdunkorchid 

please add order by

supler:
@slamdunkorchid 

please go ahead,thanks

PR #116, opened 2017-02-16
[CBRD-20928] add cases for the fix of server crash

slamdunkorchid:
http://jira.cubrid.org/browse/CBRD-20928


supler:
@slamdunkorchid 

please add order by to make resultset is stable, thanks

supler:
@slamdunkorchid 

please go ahead, thanks

PR #117, opened 2017-02-16
[CBRD-20846] add cases for the fix:Handling of CTE pointers in parse tree copy

slamdunkorchid:
http://jira.cubrid.org/browse/CBRD-20846


supler:
@slamdunkorchid 

please go ahead,thanks

PR #118, opened 2017-02-17
[CUBRIDQA-12]replace the date by a fixed value to make result stable.

jinmeihua:


supler:
@jinmeihua 

please commit your updates, thanks

PR #122, opened 2017-02-17
[CBRD-20836] add cases for the fix: Return real error for CTE name ambiguity

slamdunkorchid:
http://jira.cubrid.org/browse/CBRD-20836


supler:
@slamdunkorchid 

please go ahead,thanks

PR #123, opened 2017-02-17
[CBRD-20937] add cases for the fix:Recursive CTE optimization restriction

slamdunkorchid:
http://jira.cubrid.org/browse/CBRD-20937


supler:
@slamdunkorchid 

please go ahead, thanks

PR #124, opened 2017-02-20
[CUBRIDQA-12] drop table and add sql_by_cci answer.

jinmeihua:


supler:
@jinmeihua 

please go ahead,thanks

PR #129, opened 2017-02-20
[CBRD-20968] add cases for the fix:Recursive CTE optimization restriction, analytics

slamdunkorchid:
http://jira.cubrid.org/browse/CBRD-20968
refer to：
http://jira.cubrid.org/browse/CBRD-20957
Analytic functions must be applied on each recursive level of the CTE individually.


supler:
@slamdunkorchid 

please go ahead, thanks

PR #130, opened 2017-02-20
[CBRD-20957] add cases for the fix:Recursive CTE optimization restriction, analytics

slamdunkorchid:
http://jira.cubrid.org/browse/CBRD-20957


supler:
@slamdunkorchid 

please go ahead, thanks

PR #135, opened 2017-02-22
[CBRD-20978] Return error for cte.* usage in count function

slamdunkorchid:
http://jira.cubrid.org/browse/CBRD-20978


supler:
@slamdunkorchid 

please go ahead, thanks

PR #145, opened 2017-02-24
[CBRD-20919] update patch case

slamdunkorchid:
http://jira.cubrid.org/browse/CBRD-20919
like CUBRIDSUS-10854 ,such cases need update 8.4.1,8.4.3.8.4.4 driver patch files


supler:
@slamdunkorchid 

please go ahead, thanks

PR #148, opened 2017-03-01
[CBRD-20946] Allow only UNION ALL in recursive CTE

slamdunkorchid:
http://jira.cubrid.org/browse/CBRD-20946


supler:
@slamdunkorchid 

please go ahead, thanks

PR #149, opened 2017-03-06
[CBRD-21042] Add test case for the fix handling of error from semantic_type

jiyanping:
Add test case for the fix handling of error from semantic_type.
http://jira.cubrid.org/browse/CBRD-21042


mandy2016:
@jiyanping 
please merge it.

PR #158, opened 2017-03-15
[CBRD-21069] add cases: Handle reset of scan tuple number when a new page is fetched

slamdunkorchid:
http://jira.cubrid.org/browse/CBRD-21069


supler:
@slamdunkorchid 

I think it will be better to design more scenario for this feature, for instance, set some condition for recursive query make it less than the threshold , it should work well.

more scenario with the different condition such as where condition, having, limit ....

slamdunkorchid:
@supler 
I have expend cases, it is enough for  CBRD-21069
more test cases will be added into CBRD-20811 which would be more better

supler:
@slamdunkorchid 

please go ahead, thanks

PR #160, opened 2017-03-16
[CBRD-21077] add cases:Return error if CTE max recursion is reached when list optimization is disabled

slamdunkorchid:
http://jira.cubrid.org/browse/CBRD-21077


supler:
@slamdunkorchid 

please go ahead, thanks

PR #161, opened 2017-03-17
[CBRD-21083]This test case blocks sql regression test. Remove it temporarily.

jinmeihua:
http://jira.cubrid.org/browse/CBRD-21083

Test case "cbrd_21077.sql" blocks the sql daily regression test. 
Remove it from the repository temporarily and move back it when CBRD-21083 fixed.


fanzaiqiang:
@jinmeihua 

Please go ahead to remove it. Thank you.

PR #164, opened 2017-03-22
[CBRD-21084]reset the system param CTE_MAX_RECURSIONS to default value.

jinmeihua:
http://jira.cubrid.org/browse/CBRD-21083 : replace the removed test case.
http://jira.cubrid.org/browse/CBRD-21084 : reset the system param CTE_MAX_RECURSIONS to default value.


sql/_13_issues/_17_1h/cases/cbrd_21069.sql
```diff
@@ -250,3 +250,4 @@ FROM MyCTE AS R where R.Number>2000000
 )
 SELECT Number
 FROM MyCTE;
+set system parameters 'CTE_MAX_RECURSIONS=2000';
```
eseokoh: It should be `DEFAULT`, not 2000.

PR #175, opened 2017-04-06
[CBRD-20758]revise TCs ans answers for 'Update the latest timezone database from IANA'

slamdunkorchid:
anser changed，need generate new patch files


supler:
@slamdunkorchid 

please go ahead, thanks

PR #182, opened 2017-04-17
[CBRD-20158] update patch file,low version not support enum type and binary

slamdunkorchid:
update patch files .


supler:
@slamdunkorchid 

please go ahead, thanks

PR #247, opened 2017-10-10
 [CUBRIDQA-253]remove/modify duplicated test cases.

jinmeihua:


jinmeihua:
Passed the sql regression test.

PR #253, opened 2017-10-16
[CUBRIDQA-387]Add candidate answer.

jinmeihua:


eseokoh:
I don't think it is a right choice.

jinmeihua:
Change queries for deterministic results instead of add candidate answer .
See https://github.com/CUBRID/cubrid-testcases/pull/255.

PR #255, opened 2017-10-17
[CUBRIDQA-387]SELECT blocks another SELECT does not make sense. Remove these test cases.

jinmeihua:
http://jira.cubrid.org/browse/CUBRIDQA-387


isolation/_02_RepeatableRead/function/counter_function/select_select_rownum_02.ctl
```diff
@@ -40,9 +40,9 @@ C1: COMMIT WORK;
 MC: wait until C1 ready;
 
 /* test case */
-C1: SELECT * FROM t1 WHERE ROWNUM >= 3 and ROWNUM <= 4;
+C1: SELECT * FROM (select * from t1 order by id) as t WHERE ROWNUM >= 3 and ROWNUM <= 4;
```
eseokoh: Why don't you simply use?
```sql
SELECT * FROM t1 ORDER BY id LIMIT 3, 2;
```

jinmeihua: Because the name of the case is 'select_select_rownum_02.ctl', it intened to test with ROWNUM.
So I keep the query with ROWNUM.

eseokoh: In fact, the case is useless from the point of MVCC.
it does not make sense that SELECT blocks another SELECT. 
it is better to keep the case efficient (or remove it).

jinmeihua: There are several similar tests are the case, 
- https://github.com/cubrid/cubrid-testcases/blob/develop/isolation/_01_ReadCommitted/function/counter_function/select_select_rownum_01.ctl
- https://github.com/cubrid/cubrid-testcases/blob/develop/isolation/_01_ReadCommitted/function/counter_function/select_select_rownum_01_1.ctl
- https://github.com/cubrid/cubrid-testcases/blob/develop/isolation/_01_ReadCommitted/function/counter_function/select_select_rownum_02.ctl
- https://github.com/cubrid/cubrid-testcases/blob/develop/isolation/_02_RepeatableRead/function/counter_function/select_select_rownum_01.ctl
- https://github.com/cubrid/cubrid-testcases/blob/develop/isolation/_02_RepeatableRead/function/counter_function/select_select_rownum_02.ctl (the test case in this PR)
- https://github.com/cubrid/cubrid-testcases/blob/develop/isolation/_05_ReadCommitted_RepeatableRead/function/counter_function/select_select_rownum_01.ctl
- https://github.com/cubrid/cubrid-testcases/blob/develop/isolation/_05_ReadCommitted_RepeatableRead/function/counter_function/select_select_rownum_02.ctl

should we remove all these ?

fanzaiqiang: I checked each test case in above list.  They're similar on ROWNUM and not effective on MVCC test and hard to stabilize. It's better to delete them.

PR #269, opened 2017-10-26
[CBRD-21522] JSON Type Tests Update/Insert

paulgheorghecristian:
http://jira.cubrid.org/browse/CBRD-21522

I have updated one test regarding the system table "_db_domain" to not fail anymore. This is where the QA team should put the new JSON tests.


eseokoh:
I'm sorry patch does not help.
When we accept this change before merging JSON patch, CI and other PRs will fail.

one solution is to remove this test case, since it does not look helpful.
I will ask QA to look into the case.

eseokoh:
@fanzaiqiang 

Please investigate the case. I don't see any useful test point. 
I propose to remove the case.

fanzaiqiang:
@eseokoh 

Got it. We will investigate the test case.

PR #420, opened 2018-04-02
[CUBRIDQA-703]enhance MC actions and add order by to make result stable.

jinmeihua:


isolation/_05_ReadCommitted_RepeatableRead/no_index_column/aggregate/insert_select_01_1.ctl
```diff
@@ -36,13 +36,15 @@ MC: wait until C1 ready;
 
 /* test case */
 C1: update t set id=col-1 where mod(id,2)=0;
+MC: wait until C2 ready;
```
slamdunkorchid: wait until C2 ready; 
to be
wait until C1 ready;

isolation/_05_ReadCommitted_RepeatableRead/no_index_column/aggregate/insert_select_01_1.ctl
```diff
@@ -36,13 +36,15 @@ MC: wait until C1 ready;
...
 C2: insert into t select t1.* from (select t.* from t order by id) t1 where mod(id,3)=0;
+MC: wait until C3 ready;
```
slamdunkorchid: the same as above

isolation/_05_ReadCommitted_RepeatableRead/no_index_column/aggregate/insert_select_01_1.ctl
jinmeihua: It is intended.
The test case is expect to show C3's result first , then C2 , then C1.

isolation/_05_ReadCommitted_RepeatableRead/no_index_column/aggregate/insert_select_01_1.ctl
jinmeihua: Same as above.

isolation/_05_ReadCommitted_RepeatableRead/no_index_column/aggregate/insert_select_01_1.ctl
jinmeihua: I think two ways are right.
What you said may be easier to understand.
I will modify it.

PR #424, opened 2018-04-12
[CUBRIDQA-20]remove sleep() and refine MC actions.

jinmeihua:


fanzaiqiang:
We made the change for the reason of http://jira.cubrid.org/browse/CBRD-22000.
Ensure to have a stable order for insertion.

PR #439, opened 2018-05-04
[CUBRIDQA-20]remove sleep() function to make test case stable.

jinmeihua:


fanzaiqiang:
Okay.  Make C2 executes before C1 as improvement.
With old behavior, sleep() make C1 executes delayed.

PR #456, opened 2018-06-06
[CBRD-22118]update statistics statements authorization changed.

jinmeihua:
Revise required by http://jira.cubrid.org/browse/CBRD-22118


sql/_07_misc/_01_statistics/answers/1031.answer
```diff
@@ -13,11 +13,13 @@ db_user
 null     
 
 ===================================================
-0
+Error:-158
```
slamdunkorchid: you may change sql cases,to avoid error code,such as

call add_user('test_user') on class db_user;

create class t1(c1 int);

create class t2(c1 int);

create class t3(c1 int);
grant alter on t1,t2,t3 to test_user; 

call login('test_user') on class db_user;

update statistics on t1, t2, t3;

call login('dba') on class db_user;

drop class t1, t2, t3;

call drop_user('test_user') on class db_user;

sql/_07_misc/_01_statistics/answers/1033.answer
```diff
@@ -13,11 +13,13 @@ db_user
 null     
 
 ===================================================
-0
+Error:-158
```
slamdunkorchid: the same as above

PR #459, opened 2018-06-13
[CUBRIDQA-505] unstable,related to current datetime,so delete it

slamdunkorchid:
select to_timestamp_tz('1999 Tuesday -10', 'YYYY DAY TZH'); 

it is unstable,related to current's month,so delete this code


jinmeihua:
There is an open issue regarding this test case.
See http://jira.cubrid.org/browse/CBRD-20292 . 
I think we can approve this PR at first and add it back when the issue resolved.

@fanzaiqiang 
What do you think?

PR #515, opened 2018-11-02
[CBRD-22419] add test cases: null key values,comment.query plan for online index

slamdunkorchid:
refer to http://jira.cubrid.org/browse/CBRD-22419


slamdunkorchid:
mistake,close it

PR #522, opened 2018-11-05
[CUBRIDQA-807][CBRD-21506] add test cases:'create index ...with online' grammar

slamdunkorchid:
refer to http://jira.cubrid.org/browse/CBRD-21506


sql/_32_online_index/_001_index/cases/1002.sql
```diff
@@ -0,0 +1,24 @@
...
+);
+alter class ddl_0001 drop constraint "u_ddl_0001_col6" ;
+alter class ddl_0001 add unique u_idx(col1,col2,col3) with online ;
+create index idx_col1 on ddl_0001(col1) ;
+
+
+
+
+drop class ddl_0001;
+rollback;
```
fanzaiqiang: rollback can be removed since it is auto-commit.

Also check other cases.

PR #582, opened 2018-12-13
[CBRD-22408] add test cases: session variable usage in create/insert/replace/delete/update statements

slamdunkorchid:


slamdunkorchid:
refer to http://jira.cubrid.org/browse/CBRD-22408

PR #606, opened 2019-01-04
[CUBRIDQA-12]make result stable.

jinmeihua:
* fix inserting value.
* add ordered hint


sql/_31_cherry/issue_22161_CTE_extensions/_04_replace/answers/cast_time.answer
```diff
@@ -90,7 +90,7 @@ var_i    j
 1
 ===================================================
 i    j    
-04/20/2018     null     
+04/20/2019     null
```
slamdunkorchid: please also change the insert value 420 to 050420

PR #613, opened 2019-01-15
[CBRD-22677] add test case for this crash issue

mandy2016:
http://jira.cubrid.org/browse/CBRD-22677


sql/_31_cherry/issue_22162_more_json_functions/cases/cbrd_22677.sql
```diff
@@ -0,0 +1,2 @@
+SET @j = '["abc", [{"k": "10"}, "def"], {"x":"abc"}, {"y":"bcd"}]';
+SELECT JSON_SEARCH(@j, 'all', '%b%', NULL, '$[3].');
```
jinmeihua: Please drop variable at the end of test case.
drop variable @j ;

PR #644, opened 2019-03-25
[CBRD-22054] modify cases,avoid nok and make it stable

slamdunkorchid:
avoid to save nok result and make it stable
modify cases


slamdunkorchid:
refer to https://github.com/CUBRID/cubrid-testcases/pull/625

PR #681, opened 2019-07-25
[CUBRIDQA-12]Added separator due to incorrect answer submission

Zhaojia2019:
Modify the answer.


fanzaiqiang:
Please refine title. Do not use "modify some file" which only show your action.
Improve like "add line sepeartor due to xxx"

PR #705, opened 2019-09-30
[CUBRIDQA-911] add test case

slamdunkorchid:
refer to http://jira.cubrid.org/browse/CUBRIDQA-911


slamdunkorchid:
the test cases need be updated later, now add into develop first

PR #713, opened 2019-10-14
[CUBRIDQA-911] add test cases parallel index loader

slamdunkorchid:
refer to http://jira.cubrid.org/browse/CUBRIDQA-911


fanzaiqiang:
Please drop csql.err and then merge codes.

PR #720, opened 2019-10-18
[CBRD-23186] Add test case for this issue

mandy2016:
http://jira.cubrid.org/browse/CBRD-23186


sql/_13_issues/_19_2h/cases/cbrd_23186.sql
```diff
@@ -0,0 +1,7 @@
+drop table if exists t;
+create table t (a string, b string);
+insert into t values ('c',null);
+create index idx2 on t(nvl(a,b)) invisible;
+show index from t;
+commit;
+update t set a = null;
```
jinmeihua: Please add the following queries in the end. 
```
select * from t; 
drop table if exists t;
```

PR #726, opened 2019-10-23
[CBRD-22162] Add some i18n test cases for json test

mandy2016:


fanzaiqiang:
It's not priority to submit json with i18n cases.

PR #729, opened 2019-10-29
[CBRD-23319]add about core happen test

Zhaojia2019:
http://jira.cubrid.org/browse/CBRD-23319


Zhaojia2019:
add shell case ,close this.

PR #734, opened 2019-10-31
[CBRD-22714]add about core happen test

Zhaojia2019:
http://jira.cubrid.org/browse/CBRD-22714


sql/_13_issues/_19_2h/cases/cbrd_22714.sql
```diff
@@ -0,0 +1,12 @@
...
+
+create table t1 (id int primary key, title varchar(100));
+
+create table t2 as
+  WITH RECURSIVE cte AS ( 
+    SELECT id, title FROM t1
+    UNION ALL 
+    SELECT id + 1, 'test title' FROM cte where id<10
+  ) SELECT  * FROM cte;
+
```
jinmeihua: I think it would be better to add queries like 
```
select * from t2 order by ...
```
,
```
  WITH RECURSIVE cte AS ( 
    SELECT id, title FROM t1
    UNION ALL 
    SELECT id + 1, 'test title' FROM cte where id<10
  ) SELECT  * FROM cte order by ...;
```
and so on, to check the result.

PR #822, opened 2020-08-04
[CBRD-22925] add new case for this issue

ssihil:
http://jira.cubrid.org/browse/CBRD-22925


sql/_13_issues/_20_2h/cases/cbrd_22925.sql
```diff
@@ -0,0 +1,20 @@
...
+
+CREATE TABLE tmp1(col1 INT, col2 INT, col3 int);
+CREATE INDEX idx ON tmp1(col1,col2);
+
+select 1
+  from tmp1
+ where col3 = 1
+   and col3 <> 2
+   and col2 in (?,?)
+   and col1 in ('1','2');
```
shparkcubrid, line 11: How about binding host variable values using 'prepare' and 'execute' statements?

PR #888, opened 2020-10-19
[CBRD-23686] Add test case for if the number of index columns is large, the optimizer cannot select a more appropriate index.

sy052:
Please refer to [http://jira.cubrid.org/browse/CBRD-23686](http://jira.cubrid.org/browse/CBRD-23686)


ssihil (approved):

sql/_13_issues/_20_2h/cases/cbrd_23686.queryPlan
```diff
@@ -0,0 +1 @@
+
```
shparkcubrid, line 1: redundant

shparkcubrid (approved):

PR #889, opened 2020-10-21
[CBRD-23599] Add new test case for Error occurs select list has in the order of an scalar subquery and expression which include an analytic function.

sy052:
Please refer to [http://jira.cubrid.org/browse/CBRD-23599](http://jira.cubrid.org/browse/CBRD-23599)


ssihil (approved):

ssihil (approved):

hgryoo:
I don't know about the exact convention of the test cases. But IMO, Table and column names need to be given unmeaningful names such as t1, a1, a2 If I refer to other test cases.

hgryoo (approved):

PR #920, opened 2020-12-04
[CBRD-23767] add cci answer.

tw-kang:
http://jira.cubrid.org/browse/CBRD-23767


tw-kang:
verified as a test case error.

PR #965, opened 2021-06-28
[CBRD-23816] modified answer for support HASH TEMP FILE

tw-kang:
http://jira.cubrid.org/browse/CBRD-23816

### DIFF
--- a/sql/_13_issues/_20_2h/answers/cbrd_23665.answer_cci
+++ b/sql/_13_issues/_20_2h/answers/cbrd_23665.answer_cci
@@ -31,7 +31,7 @@ Query Plan:
 Trace Statistics:
   SELECT (time: ?, fetch: ?, ioread: ?)
     SCAN (table: tmp_hls), (heap time: ?, fetch: ?, ioread: ?, readrows: ?, rows: ?)
-      SCAN (hash temp, build time: ?, time: ?, fetch: ?, ioread: ?, readrows: ?, rows: ?)
+      SCAN (hash temp(m), build time: ?, time: ?, fetch: ?, ioread: ?, readrows: ?, rows: ?)
     SUBQUERY (uncorrelated)
       SELECT (time: ?, fetch: ?, ioread: ?)
         SCAN (table: tmp_hls), (heap time: ?, fetch: ?, ioread: ?, readrows: ?, rows: ?)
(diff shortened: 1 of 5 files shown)


sql/_13_issues/_20_2h/answers/cbrd_23665.answer_cci
```diff
@@ -949,7 +949,7 @@ trace
 Trace Statistics:
   SELECT (time: ?, fetch: ?, ioread: ?)
     SCAN (temp time: ?, fetch: ?, ioread: ?, readrows: ?, rows: ?)
-      SCAN (hash temp(h), build time: ?, time: ?, fetch: ?, ioread: ?, readrows: ?, rows: ?)
+      SCAN (hash temp(f), build time: ?, time: ?, fetch: ?, ioread: ?, readrows: ?, rows: ?)
```
shparkcubrid: It is necessary to check whether the answer sheet should be changed or the test case should be changed.

sql/_13_issues/_20_2h/answers/cbrd_23665.answer_cci
```diff
@@ -913,7 +913,7 @@ trace
 Trace Statistics:
   SELECT (time: ?, fetch: ?, ioread: ?)
     SCAN (temp time: ?, fetch: ?, ioread: ?, readrows: ?, rows: ?)
-      SCAN (hash temp(h), build time: ?, time: ?, fetch: ?, ioread: ?, readrows: ?, rows: ?)
+      SCAN (hash temp(m), build time: ?, time: ?, fetch: ?, ioread: ?, readrows: ?, rows: ?)
```
shparkcubrid: It is necessary to check whether the answer sheet should be changed or the test case should be changed.

tw-kang: in the testcase seems to check memory limit.
case line is 
select /*+ ordered */ count(*) from (select * from t1 limit 1) a, (select * from t1 limit 100000) b where a.a = b.a and a.b = b.b and a.c = b.c;

does this statement affect memory limit?

sql/_13_issues/_20_2h/answers/cbrd_23665.answer_cci
tw-kang: before was setting system parameter 'max_hash_list_scan_size=4M'.
now default increased to 8M.
does that mean size minimum value starts from 8M? if yes test case have to be modified.

ssihil (approved):

PR #1006, opened 2021-10-25
[CBRD-24063] Add testcase for error when Connection.close() is called multiple times in server-side JDBC

zionyun:
refer to http://jira.cubrid.org/browse/CBRD-24063


sql/_08_javasp/cases/cbrd_24063.sql
```diff
@@ -0,0 +1,6 @@
+CREATE FUNCTION sp() RETURN string as language java name 'SpTest8.SP() return java.lang.String';
+CREATE FUNCTION sp2() RETURN string as language java name 'SpTest8.SP2() return java.lang.String';
```
hgryoo: It would be better to use `CREATE OR REPLACE FUNCTION` for test cases to make stable

hgryoo (approved):

sql/_08_javasp/cases/cbrd_24063.sql
zionyun: Reflected the review. thank you

ssihil (approved):

sy052 (approved):

PR #1029, opened 2022-01-05
[CBRD-24148] Add new test case for Core dump occur when using hash list scan.

tw-kang:
http://jira.cubrid.org/browse/CBRD-24148


shparkcubrid:
How about reducing data to speed up testing? This is, of course, the condition in which the core occurs.

tw-kang:
> How about reducing data to speed up testing? This is, of course, the condition in which the core occurs.

to optimize the testing time reduce input data size 1,000,000(as-is) 410,000(to-be).

shparkcubrid (approved):

ssihil (approved):

sy052 (approved):

PR #1050, opened 2022-01-25
[CBRD-24149] Backport from develop to 11.1

swi0110:
Refer to http://jira.cubrid.org/browse/CBRD-24149, #1048 

The string-type comparison policy is changed as follows as ORACLE:
- the trailing spaces are ignored only at char-type to char-type comparison.
- the comparison with char-type and varchar-type is based on varchar-type.
- all constant string is char-type regardless it has trailing space.


swi0110:
The CBRD-24149 does not apply to 11.1

PR #1058, opened 2022-01-26
[CBRD-24141]modified answer for error code change

tw-kang:
http://jira.cubrid.org/browse/CBRD-24141


ctshim:
For the keywords marked in red in the comments of [CBRD-24141], it seems to be necessary to add a TC to check whether the query is executed successfully with the table name or column name, etc.
Please also refer to the CBRD-24141-testcase.txt file attached to the issue

ctshim (approved):

tw-kang:
> For the keywords marked in red in the comments of [CBRD-24141], it seems to be necessary to add a TC to check whether the query is executed successfully with the table name or column name, etc. Please also refer to the CBRD-24141-testcase.txt file attached to the issue

New test case will be added after this PR.

PR #1077, opened 2022-02-18
[CBRD-24083] Add new test case for conditional clause composed only of logical operations.

sy052:
refer to http://jira.cubrid.org/browse/CBRD-24083


ctshim:
Please add the case below.
```
select (1 is true);
select (1 is false);
select (0 is true);
select (0 is false);
select (2 is true);
select (2 is false);

select (1 = true);
select (true = 1);
select (1 = false);
select (false = 1);

select (0 = true);
select (true = 0);
select (0 = false);
select (false = 0);

select (2 = true);
select (true = 2);
select (2 = false);
select (false = 2);

select (1 = null), (0 = null), (null = 1), (null = 0);
select (1 is null), (0 is null);

select (null = null);
select (unknown = unknown);

select (unknown = null);
select (null = unknown);

select (null is null);
select (unknown is null);

select (null is unknown);
select (unknown is unknown); 
```

ctshim (approved):

ssihil (approved):

PR #1105, opened 2022-03-08
[CBRD-24033] Revised answers for changing the error code.

sy052:
refer to http://jira.cubrid.org/browse/CBRD-24033


youngjinj:
The CBRD-24033 issue does not require and should not be modified in the test scenario.

QA scenario change was requested only in issue CBRD-24195.
**In the CBRD-24033 issue, no QA scenario change is required.**

The reason for changing the test scenario results in this PR is because there was a change in the delete_name rule in the CBRD-24033 issue.
However, the change in the delete_name rule was reverted in the CBRD-24195 issue.
Therefore, in the CBRD-24033 issue, the test scenario change is not required and should not be changed.

Please refer to the comments below.
* http://jira.cubrid.org/browse/CBRD-24033?focusedCommentId=4760490&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4760490

youngjinj (approved):

youngjinj:
In view of this issue alone, it is correct that the answer should be changed.

zionyun:
Revert for a while to merge other issues.

PR #1107, opened 2022-03-10
[CBRD-23844] Revised sql answers for user schema

zionyun:
http://jira.cubrid.org/browse/CBRD-24195


youngjinj (approved):

medium/_01_fixed/answers/f_agg1.answer
```diff
@@ -1,5 +1,5 @@
 ===================================================
-class faculty1    max(ssn)    
+class joe.faculty1    max(ssn)    
 null     55
```
youngjinj, line 3: It should be the result below.
```
  class joe.faculty1       max(ssn)
===================================
  joe.faculty1                   55
```

youngjinj, line 3: When I execute the query, the result is not null, so why is the result of TC null?
```
  class joe.faculty1       max(ssn)
===================================
  joe.faculty1                   55
```
Even if I execute it in develop, the result of TC is null.

medium/_01_fixed/cases/fview1.sql
```diff
@@ -1,17 +1,17 @@
 autocommit off;
 call login('joe') on class db_user;
-create class xxx (a int);
-insert into xxx values(6);
-create vclass xxx_v (a1 int) as select a from xxx;
+create class joe.xxx (a int);
+insert into joe.xxx values(6);
+create vclass joe.xxx_v (a1 int) as select a from joe.xxx;
```
youngjinj: No need to specify owner while creating.
Also, an owner does not need to be specified when using an owned object.
It should be changed as below.
```
create class xxx (a int);
insert into xxx values(6);
create vclass xxx_v (a1 int) as select a from xxx;
```

medium/_01_fixed/cases/fview11.sql
```diff
@@ -1,11 +1,11 @@
 autocommit off;
 select product_code, descr, price
-from all inventory_v i
+from all joe.inventory_v i
 order by descr asc;
   select product.product_code, product.descr, product.price
```
youngjinj: Please match the indentation as well.

zionyun: reflected. please check.

medium/_01_fixed/cases/fview1.sql
zionyun: reflected. please check.

medium/_01_fixed/answers/f_agg1.answer
zionyun: I don't know why null occurs in that TC. When first written, it is expected to have a purpose.

medium/_01_fixed/cases/fview1.sql
youngjinj: Thank you.

medium/_01_fixed/cases/fview11.sql
youngjinj: Thank you.

medium/_01_fixed/answers/f_agg1.answer
youngjinj: Thank you. I hope it gets fixed later.

PR #1172, opened 2022-04-20
[CBRD-24082] Add new test cases for view merging extension

sy052:
refer to http://jira.cubrid.org/browse/CBRD-24082


ssihil (approved):

shparkcubrid:
From now on, I hope that the files will not be divided by case. You just need to specify the cases in the comments.

shparkcubrid (approved):

PR #1183, opened 2022-04-22
[CBRD-24182] Add new test cases for removal unnecessary table

sy052:
refer to http://jira.cubrid.org/browse/CBRD-24182


ssihil (approved):

shparkcubrid:
Why are you splitting each case into a file? It would be difficult to manage and there would be a lot of unnecessary create inserts. Anyway, I approve.

shparkcubrid (approved):

sy052:
> Why are you splitting each case into a file? It would be difficult to manage and there would be a lot of unnecessary create inserts. Anyway, I approve.

If the TC result needs to be revised, it's difficult to confirm, so I divided it.

PR #1248, opened 2022-07-05
[CBRD-24396] Revised cci answers for ignoring hints if invalid index name exists

sy052:
refer to http://jira.cubrid.org/browse/CBRD-24396


ctshim (approved):

ctshim:
For your information, nchar and varnchar have not been supported since v9.0. How about removing the column from the QA scenario?

sy052:
> For your information, nchar and varnchar have not been supported since v9.0. How about removing the column from the QA scenario?

Thank you for the advice. 
 I deleted that one because it will be fixed another issue

ssihil (approved):

PR #1263, opened 2022-07-19
[CBRD-24424] Add new test case for core dump in the show statement

zionyun:
refer to http://jira.cubrid.org/browse/CBRD-24424


ssihil (approved):

sql/_13_issues/_22_2h/cases/cbrd_24424.sql
```diff
@@ -0,0 +1,33 @@
+show heap header of dba._db_class;
```
youngjinj, line 33: The test case does not contain the comments I made.

If this test case fails during development, we need to check what the test case is for.
If possible, please include comments as well.

zionyun: Okay. Added comments.

youngjinj (approved):

PR #1291, opened 2022-09-06
[CBRD-24388]Add new testcase for detailed error contents when generating foreign keys.

tw-kang:
http://jira.cubrid.org/browse/CBRD-24388


tw-kang:
This case will be written in shell test form.

PR #1318, opened 2022-10-15
[CBRD-24473] restore optimization level before exec drop queries

kisoo-han:
http://jira.cubrid.org/browse/CBRD-24473

In addition to #1317, 
* restore optimization level to previous one
* restore optimization level before 'drop' queries.
   (opt level 514 will not execute queries)


zionyun:
@kisoo-han I am sorry to have bothered you. And thank you for fixing the testcase.

PR #1337, opened 2022-11-11
[CBRD-24502] Revised isolation answer for CBRD-24502

zionyun:
refer to http://jira.cubrid.org/browse/CBRD-24502


zionyun:
Wrong modification. I will create a new pr.

PR #1358, opened 2023-01-04
[CBRD-24598] Add new test case for memory allocation.

tw-kang:
http://jira.cubrid.org/browse/CBRD-24598

This issue only verified the core dump problem.
But now the order of executing the 'execute q... ' query occurs an error caused by the executing order.
ex ) 
execute q using 'A'; /* success */
execute q using 1; /* success */

execute q using 'A'; /* fail - ERROR: Cannot coerce a value of domain "character" to domain "double". */

This will be discussed and fixed in another issue. 
Until fix the Actual result will be /* ERROR -181 : Cannot coerce value of domain "character" to domain "double". */.
It will be the answer sheet and after the issue is solved it must be revised as success result.


ssihil (approved):

sql/_13_issues/_23_1h/cases/cbrd_24598.sql
```diff
@@ -0,0 +1,19 @@
...
+drop table if exists t;
+
+-- prepare required 
+prepare q from 'select decode (?, '''', c, NULL, c, -1) from table ({1}) as t (c)'; 
+
+-- success
+execute q using 'A'; 
+
+-- success
+execute q using 1;
```
youngjinj, line 15: Please check the query below once more.
```
-- success
execute q using 'A';
```
The error code will be the answer.

I would like to be able to check whether this problem has been resolved through a test case fail.

ssihil (approved):

sql/_13_issues/_23_1h/cases/cbrd_24598.sql
tw-kang: opposite case and repeated execution case is added.

sql/_13_issues/_23_1h/cases/cbrd_24598.sql
```diff
@@ -0,0 +1,37 @@
...
+-- error : temporary
+-- known bug
+execute p using 1;
+
+-- success
+execute p using 'A';
+
+-- error : temporary
+-- known bug
+execute p using 1;
```
youngjinj: The below seems to be correct.
```suggestion
-- success
execute p using 1;

-- error : temporary
-- known bug
execute p using 'A';

-- success
execute p using 1;
```

sql/_13_issues/_23_1h/answers/cbrd_24598.answer
```diff
@@ -0,0 +1,32 @@
...
+decode( ?:0 , '', c, null, c, -1)    
+-1     
+
+===================================================
+decode( ?:0 , '', c, null, c, -1)    
+-1     
+
+===================================================
+decode( ?:0 , '', c, null, c, -1)    
+-1
```
youngjinj: An error occurs and it fails.
Shouldn't the error code be included in the answer?

sql/_13_issues/_23_1h/answers/cbrd_24598.answer
```diff
@@ -0,0 +1,32 @@
...
+
+===================================================
+0
+===================================================
+decode( ?:0 , '', c, null, c, 'Z')
+Z
+
+===================================================
+decode( ?:0 , '', c, null, c, 'Z')
+Z
```
youngjinj: An error occurs and it fails.
Shouldn't the error code be included in the answer?

kwonhoil (approved):

sql/_13_issues/_23_1h/answers/cbrd_24598.answer
tw-kang: The answer sheet is written with the expected result.
Now the query executing order problem is detected as a known bug.
So this will be detected as a fail and verified as a known bug.
Until the bug is solved we will track this as a failure case in the regression test.

youngjinj: Fixing this bug will take longer than expected.

Until then, the sql test will continue to fail.
As a result, development teams can't merge code.

What I said I want to check through fail is not to keep it as fail.
Right now, it makes an answer with the actual result, and expects this TC to fail when this bug is fixed.

Again, if it continues to fail, it will be difficult to merge the code.

tw-kang: I missed the circle CI sql test. 
The answer sheet is revised to occur errors and it will be revised after the known bug is solved.

youngjinj (approved):

PR #1363, opened 2023-01-10
[CBRD-24563]revised answer refer to library change

tw-kang:
http://jira.cubrid.org/browse/CBRD-24563


hgryoo (changes requested):
Since we are still supporting the cppstd, the existing regression tests should not be erased. In the 'specification changes' section in JIRA, I wrote that 'It is necessary to copy the existing regression tests for the new library backend RE2 and modify the answers if required.'.

hgryoo:
To avoid regression failures in sql CI, It is enough to set the 'regexp_engine' system parameters need be set to 'cppstd'. Then new test cases for the re2 library need to be added.

sql/_32_damson/cbrd_22783_REGEXP_functions/answers/REGEXP_COUNT_unicode.answer
```diff
@@ -1,20 +1,22 @@
...
 ===================================================
+0
+===================================================
 regexp_count(_utf8'가나다가나다가나다라' collate utf8_ko_cs, _utf8'(가나)다' collate utf8_ko_cs, 1)    
 3     
 
 ===================================================
 regexp_count(_utf8'가나 가나다라 마바사아 자차카타 파하' collate utf8_ko_cs, _utf8'[[:alpha:]]+' collate utf8_ko_cs, 1)    
-5     
+0
```
tw-kang: @hgryoo 
In Unicode included cases the execution result seems 'regexp_engine' parameter is not set as 'cppstd'. The result is the same as 're2'.

hgryoo (approved):

ssihil (approved):

PR #1426, opened 2023-03-28
[CBRD-24725] Add test case for Core occurs in aggregation query

zionyun:
refer to http://jira.cubrid.org/browse/CBRD-24725


zionyun:
This test case must be merged into 10.1 branch.

PR #1457, opened 2023-04-19
[CBRD-24770] Add testcase for when a connect by query is executed wit…

ssihil:
…h covered index
refer to http://jira.cubrid.org/browse/CBRD-24770
Fix issues not resolved by http://jira.cubrid.org/browse/CBRD-24705


youngjinj:
The test cases for CBRD-24770 are the same as for CBRD-24705.
So it is more efficient to change the test cases of CBRD-24705 to those of CBRD-24770.
(From the point of view of creating as few test cases as possible)

I will approve, but please make a decision based on this content.

youngjinj (approved):

kwonhoil (approved):

PR #1528, opened 2023-08-01
[CBRD-24914] Exclude testcase caused by core dump.

tw-kang:
http://jira.cubrid.org//browse/CBRD-24914


ssihil (approved):

swi0110:
How about the add the exclude file for 11.3, not removing the test case?

tw-kang:
@swi0110 
Exclude list for sql test does not exist, so after the core is fixed this case will be included again as soon as possible.

PR #1538, opened 2023-08-08
[CBRD-24502] Backport for the problem of updating statistical information when performing DDL (11.2)

swi0110:
Refer to http://jira.cubrid.org/browse/CBRD-24502, http://jira.cubrid.org/browse/CBRD-24715, #1485 

This test case reported a bug in the CBRD-24715.
but it is not a bug & won't fix it.
It is CUBRID's design, related to the CBRD-24502.
- (http://jira.cubrid.org/browse/CBRD-24715?focusedCommentId=4764235&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4764235)

### DIFF
--- a/sql/_08_javasp/answers/case_index_scan_01.answer
+++ b/sql/_08_javasp/answers/case_index_scan_01.answer
@@ -1,0 +1,234 @@
+===================================================
+0
+===================================================
+0
+===================================================
+0
+===================================================
+0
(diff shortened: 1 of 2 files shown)


swi0110:
I mistook.
This test case is not needed on release/11.2.

PR #1629, opened 2024-03-05
[CBRD-24477] only check ISNULL for some built-in function results

hyunikn:
http://jira.cubrid.org/browse/CBRD-24477

Only check ISNULL for results of several built-in functions because

- some of them change on every run: drandom, random, utc_date, utc_time, sys_guid
- the others change depending on then running environment: current_user, system_user, user, version,


sql/_05_plcsql/_01_testspec/_02_declaration/_02_variable/_02_uninitialized/cases/02_02_02-03_normal_function.sql
```diff
@@ -37,10 +37,8 @@ call t();
 
 
 create or replace procedure t  as
-    a varchar := USER();
+    a varchar;
```
kwonhoil, line 40: I think it would be better to modify it as follows.
This is because there is no case in which the user() function is used in the declaration part.

a varchar := case isnull(USER()) when 0 then 'ok' when 1 then 'nok' end

kwonhoil (approved):

PR #1688, opened 2024-04-08
[CBRD-25246] Added a test case for Add cur_pageid, volumeid in SELECT_PAGE_INFO

junsklee:
Refer to: http://jira.cubrid.org/browse/CBRD-25246

### DIFF
--- a/sql/_35_fig_cake/cbrd_25246/cases/cbrd_25246.sql
+++ b/sql/_35_fig_cake/cbrd_25246/cases/cbrd_25246.sql
@@ -2,11 +2,11 @@
 -- Ensure that cur_pageid and volumeid columns exist in the SELECT_PAGE_INFO table
 -- when the hint /*+ select_page_info */ is used in a SELECT query
 
-drop table if exists t111;
-create table t111 (col1 varchar(20), col2 varchar(20), col3 varchar(20), col4 varchar(20), col5 numeric(20,10), col6 varchar(20), col7 varchar(20), col8 varchar(20));
-select /*+ select_page_info */ * from t111;
+drop table if exists tbl;
+create table tbl (col_a varchar(20), col_b varchar(20), col_c varchar(20), col_d varchar(20), col_e numeric(20,10), col_f varchar(20), col_g varchar(20), col_h varchar(20));
+select /*+ select_page_info */ * from tbl;
 
-insert into t111 select rownum, 1, mod(rownum,10), lpad(to_char(mod(rownum,100)),20,'0'), mod(rownum,1000), lpad(to_char(mod(rownum,10000)),20,'0'), lpad(to_char(mod(rownum,100000)),20,'0'), lpad(to_char(mod(rownum,100000)),20,'0') from db_class a, db_class b, db_class c, db_class d, db_class f, db_class g limit 100;
-select /*+ select_page_info */ * from t111;
+insert into tbl select rownum, 1, mod(rownum,10), lpad(to_char(mod(rownum,100)),20,'0'), mod(rownum,1000), lpad(to_char(mod(rownum,10000)),20,'0'), lpad(to_char(mod(rownum,100000)),20,'0'), lpad(to_char(mod(rownum,100000)),20,'0') from db_class a, db_class b, db_class c, db_class d, db_class f, db_class g limit 100;
+select /*+ select_page_info */ * from tbl;
 
-drop table t111;  
+drop table tbl;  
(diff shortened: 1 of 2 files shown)


sql/_35_fig_cake/cbrd_25246/cases/cbrd_25246.sql
```diff
@@ -0,0 +1,12 @@
...
+-- Ensure that cur_pageid and volumeid columns exist and show appropriate values in the 
+-- SELECT_PAGE_INFO table when the hint /*+ select_page_info */ is used in a SELECT query
+
+drop table if exists tbl;
+create table tbl (col_a varchar(20), col_b varchar(20), col_c varchar(20), col_d varchar(20), col_e numeric(20,10), col_f varchar(20), col_g varchar(20), col_h varchar(20));
+select /*+ select_page_info */ * from tbl;
+
+insert into tbl select rownum, 1, mod(rownum,10), lpad(to_char(mod(rownum,100)),20,'0'), mod(rownum,1000), lpad(to_char(mod(rownum,10000)),20,'0'), lpad(to_char(mod(rownum,100000)),20,'0'), lpad(to_char(mod(rownum,100000)),20,'0') from db_class a, db_class b, db_class c, db_class d, db_class f, db_class g limit 100;
+select /*+ select_page_info */ * from tbl;
+
```
kwonhoil, line 8: I think it would be a good idea to add an order by clause to prevent the answer from being changed.

junsklee: A new query has been added with an ORDER BY clause in the latest commit.

junsklee:
The test queries have been edited so that they simply verify the working state of the cur_pageid and volumeid columns in SELECT_PAGE_INFO, and no longer check for explicit values of cur_pageid, as the output of SELECT_PAGE_INFO is more than likely going to contain different values for every database server and is also subject to change (e.g., when a database is modified).

sql/_35_fig_cake/cbrd_25246/cases/cbrd_25246.sql
```diff
@@ -0,0 +1,43 @@
...
+    END AS assert__p_cur_pageid__p_next_pageid,
+    CASE 
+        WHEN p_cur_pageid - 1 = p_prev_pageid THEN 'p_cur_pageid-1=p_prev_pageid: PASS' 
+        WHEN p_prev_pageid IS NULL THEN
+            CASE 
+                WHEN p_cur_pageid IS NULL THEN 'p_prev_pageid & p_cur_pageid IS NULL: FAIL'
+                WHEN p_next_pageid IS NULL THEN 'p_prev_pageid & p_next_pageid IS NULL: FAIL'
+                ELSE 'p_prev_pageid IS NULL: PASS'
+            END
+        ELSE 'p_cur_pageid-1!=p_prev_pageid: FAIL'
```
shparkcubrid: The p_cur_volumeid may not be 0. Also, the difference between the current page and the next or prev page may not be 1. What can be guaranteed is that when next is -1, cur page has the maximum value, and when prev is NULL, it has the minimum value.

shparkcubrid (approved):

sql/_35_fig_cake/cbrd_25246/cases/cbrd_25246.sql
junsklee: The query written here is only applicable for this test case only, meaning it is not used universally across different environments and scenarios. This means for our test case environment, the volume ID should be 0,

junsklee: As seen in the earliest commits, the answer sheets are usually hard-coded with fixed values in these scenarios; however, it was found that the starting page IDs varied for every database and may vary across future updates, which is the reason the test case was rewritten in such fashion. 
This is to say that the queries written here are applicable to this TC (i.e. the query that creates and inserts 2 pages worth of values into _tbl_) only, and are not used/reused universally across different instances and scenarios. Please correct me if I'm wrong, but unless something had gone wrong in our test case environment, the volume ID should remain 0 (as no other data exists in the testing DB) and the page numbers should be incremented by a factor of one (reflecting the inserted data in tbl after its creation).

bagus-kim (approved):

tw-kang (approved):

ssihil (approved):

sql/_35_fig_cake/cbrd_25246/cases/cbrd_25246.sql
shparkcubrid: Adding only guaranteed test cases will prevent unnecessary errors from occurring later. But it's your job too, so do whatever you want.

junsklee: Understood. I have considered your suggestions and have made the following changes to the pass conditions in the latest commit:

Row returns TRUE if:
- p_cur_volumeid>=0
- p_cur_pageid>=0

kwonhoil (approved):

PR #1757, opened 2024-06-21
[CBRD-25323] Perform the 'Eliminate INNER JOIN' optimization for the first table, and remove the join type for the second table (#1720)

youngjinj:
http://jira.cubrid.org/browse/CBRD-25323

backport #1720


tw-kang:
tc 백포트는 qa 팀에서 진행함에 따라 해당 pr close 합니다.

PR #1793, opened 2024-08-01
Revert "[CBRD-25335] Add testcase for a problem where nodes without a join relationship ar…"

sjkimxxx:
Reverts CUBRID/cubrid-testcases#1781


sjkimxxx:
Review is not completed.

PR #1880, opened 2024-09-24
[CBRD-24896] Add new testcase for Problem when accesing system table/views with search condition comparing set values

tw-kang:
http://jira.cubrid.org/browse/CBRD-24896

Modified to determine that index scan is impossible in the following cases
`set = set`
`{col1,col2} = {1,2}`
fix so that only collaction types written in parentheses can index-scan.


ssihil:
jira description에 있는 재현절차 중 2), 3) (system table) 도 추가하는 것이 좋겠습니다.

ssihil (approved):

kwonhoil (approved):

shparkcubrid (approved):

PR #1959, opened 2024-11-13
[CBRD-25274] Add [bugfix] Call of SYSDATE results in a type error

kiho-um:
Refer to http://jira.cubrid.org/browse/CBRD-25274

As shown below, call of SYSDATE results in a type error. 
This is because CUBRID does not allow SYSDATE() with the parenthesis and rejecdts "select sysdate() from dual" which is used to check semantics of sysdate call.


kiho-um:
PR 잘 못 되었습니다.

PR #1960, opened 2024-11-13
[CBRD-25267] Add the output order of dbms_output.put_line is strange.

kiho-um:
Refer to http://jira.cubrid.org/browse/CBRD-25267

The output order of dbms_output.put_line is strange.


kiho-um:
pr 잘 못 했습니다.

PR #2012, opened 2024-12-27
[CBRD-25266] (bugfix) CSQL core dump when an undefined column name in a Static SQL Insert statement

kiho-um:
Refer to http://jira.cubrid.org/browse/CBRD-25266

The column name undefined use in a Static SQL Insert statement.
error code : [err] Stored procedure compile error: does not have attribute j


kiho-um:
merge가 "feature/plcsql-p1"로 되어 있어서 다시 develop로 PR를 했습니다.

kwonhoil (approved):

PR #2041, opened 2025-01-10
[CBRD-25800] Revised TC to reflect changes in SHOW GRANTS - [user_schema] is now printed in lowercase instead of uppercase.

junsklee:
Refer to http://jira.cubrid.org//browse/CBRD-25800


junsklee:
https://github.com/CUBRID/cubrid-testcases/pull/2037 과 중복되어 close.

PR #2257, opened 2025-07-22
[CBRD-25611] Add a sql test case for using GROUP BY with ROLLUP causes the expression to output NULL

kimhak-tech:
Refer to http://jira.cubrid.org/browse/CBRD-25611 

**Issue:**
When using `GROUP BY` with `ROLLUP`, if an expression  or user-defined function (stored procedure) is used in the preceding condition of `GROUP BY` and a column in the subsequent condition is referenced, it is displayed as `NULL`.

**Test Coverage**
Following test scenarios are being cover:
- Different data type (int, float, double)
- Different insertion value (int, float, negative, null)
- Use various built-in functions as expressions (ceil, floor, round, abs)
- Column reference variations with built-in function(ceil)
   - Column present in both `SELECT` and `GROUP BY`
   - Column missing from `GROUP BY`
   - Column missing from both `SELECT` and `GROUP BY`
   - Different order of columns
- Column reference variations with user-defined function (same with column reference variations with built-in, but with user-defined function)
- Use nested functions as the expression
   - Built-in inside Built-in
   - Built-in inside User-Defined function
- Use multiple expressions
   - Two Built-in functions as the expression
   - Built-in and User-Defined function as the expression
- Aggregate functions
  - Built-in with Aggregate functions
  - User-defined with Aggregate functions
  - Multiple functions with different aggregate functions
- Other Cases
   - 3-Level ROLLUP
   - GROUP BY with HAVING
   - Inline Function


junsklee:
The .answer and .answer_cci files are identitcal. You only need the latter when the two drivers (JDBC vs CCI) output different answers. 

You can try it out for yourself. Remove the .answer_cci file and try re-running the test case via sql_by_cci; the test should still pass.

junsklee:
All SELECT statements should include an "ORDER BY" clause, as the order of returned rows is not guaranteed to match the order in which they were inserted. This is especially important for tables containing more than one row.

junsklee:
Also I would suggest adding in an aggregate. Right now we only project the grouping keys, so a subtle bug could still pass if the row counts are off (i.e. incorrect row membership). I'd suggest tacking on `count(*) AS cnt` to every SELECT statement (and ordering by it).

sql/_13_issues/_25_2h/cases/cbrd_25611.sql
```diff
@@ -0,0 +1,146 @@
...
+ *  Different Built-in Functions
+ *  1. Ceil
+ *  2. Floor
+ *  3. Round
+ *  4. ABS
+ *  Case with Built-in Function (Ceil)
+ *  5. original column referenced by the expression exists in both select-list and group by clause
+ *  6. original column referenced by the expression is not specified in group by clause
+ *  7. original column referenced by the expression is not specified in select-list and group by clause
+ *  8. dimension level of the original column referenced by the expression is less than that of the expression => Different Column Orders
```
junsklee: Are the columns actually shuffled? I'm not entirely sure if the column order actually varies.

sql/_13_issues/_25_2h/cases/cbrd_25611.sql
```diff
@@ -0,0 +1,146 @@
...
+evaluate('14. Built-in inside User-Defined Function');
+select /*+ recompile */ test_func(abs(c1)) AS c0, c1
+from t1 
+group by c0, c1 with rollup;
+
+-- =====================================================
+-- Multiple Expressions
+-- =====================================================
+evaluate('15. Two Built-in function as the expression');
+select /*+ recompile */ ceil(c1) as c0, floor(c2) AS c5
```
junsklee: Small nitpick (not incorrect per se, but lacks consistency/intuitiveness): Is there a particular reason why floor(c2) has an alias _c5_? My guess is to differentiate it as the alias of a column that uses a built-in function, but even then, a better name would be something like _c2_bi_.

junsklee:
Here are some edge cases worth considering:

- NULL row and a couple of negatives—ROLLUP had a habit of miscounting or NULLing things there
- 3‑level roll‑up (c0, c5, c1) just to make sure the dimension loop can’t regress
- GROUP BY + HAVING
- ORDER BY + ROLLUP—we had that weird NULL‑sorting issue mentioned in the JIRA issue comment (http://jira.cubrid.org/browse/CBRD-25611?focusedCommentId=4771291&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4771291)

sql/_13_issues/_25_2h/cases/cbrd_25611.sql
kimhak-tech: The existen of original column is being shuffled between the select statement and group by clause. 
Order changed of the column is impleted on test case 8 (dimension level of the original column).

sql/_13_issues/_25_2h/cases/cbrd_25611.sql
kimhak-tech: thanks. i'll update it.

kimhak-tech:
> Here are some edge cases worth considering:
> 
> * NULL row and a couple of negatives—ROLLUP had a habit of miscounting or NULLing things there
> * 3‑level roll‑up (c0, c5, c1) just to make sure the dimension loop can’t regress
> * GROUP BY + HAVING
> * ORDER BY + ROLLUP—we had that weird NULL‑sorting issue mentioned in the JIRA issue comment (http://jira.cubrid.org/browse/CBRD-25611?focusedCommentId=4771291&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4771291)

Update: 
- Reflected above points
- Remove .answer_cci file
- Adding aggregate function (count(*))
- Adding "ORDER BY" clause

sql/_13_issues/_25_2h/cases/cbrd_25611.sql
junsklee: Ah got it. You meant you _reversed_ the position of the columns (c0, c1) in relative to tests 5~7 (c1, c0).

junsklee: Ah got it. I didn't realize you reversed the position of the columns (c0, c1) in relative to tests 5~7 (c1, c0).

sql/_13_issues/_25_2h/cases/cbrd_25611.sql
```diff
@@ -0,0 +1,186 @@
...
+ *  12. dimension level of the original column referenced by the function is less than that of the expression
+ *  Nested Function
+ *  13. Built-in inside Built-in
+ *  14. Built-in inside User-Defined Function
+ *  Multiple Expressions and Functions
+ *  15. Two Built-in functions as the expression
+ *  16. Built-in and User-Defined function as the expression
+ *  Other Cases
+ *  17. 3-Level ROLLUP
+ *  18. GROUPING BY with HAVING
```
junsklee:
```suggestion
 *  18. GROUP BY with HAVING
```

sql/_13_issues/_25_2h/cases/cbrd_25611.sql
```diff
@@ -0,0 +1,186 @@
...
+-- Other Cases
+-- =====================================================
+
+evaluate('17. 3-Level ROLLUP');
+SELECT /*+ recompile */ ceil(c1) as c1_ceil, floor(c2) as c2_floor, c1, count(*) as cnt
+FROM t1 
+GROUP BY c1_ceil, c2_floor, c1 WITH ROLLUP
+ORDER BY c1_ceil, c2_floor, c1;
+
+evaluate('18. GROUPING BY with HAVING');
```
junsklee:
```suggestion
evaluate('18. GROUP BY with HAVING');
```

sql/_13_issues/_25_2h/cases/cbrd_25611.sql
```diff
@@ -0,0 +1,186 @@
...
+ *  14. Built-in inside User-Defined Function
+ *  Multiple Expressions and Functions
+ *  15. Two Built-in functions as the expression
+ *  16. Built-in and User-Defined function as the expression
+ *  Other Cases
+ *  17. 3-Level ROLLUP
+ *  18. GROUPING BY with HAVING
+ *  19. Inline Function
+ */
+
```
junsklee, line 35: Need to add table drop (e.g., `drop if exists t1;`) here

junsklee:
After another review of the answer file, I have come to the conclusion that for our intents and purposes in this scenario, we can go ahead and omit the `ORDER BY` from each query. This is because `GROUP BY with ROLLUP `presents null value totals in the end, but the ORDER BY messes with the ordering and puts it at the top of the result set (unless the GROUPING() function is used in the ORDER BY, but this function is unsupported in CUBRID). This causes confusion to the viewers, and since `GROUP BY` already sorts to a certain extent, we can omit the `ORDER BY` entirely for our scenario here. The created table isn't extensive, so it will be stable enough to omit the ORDER BY (which is only a problem if the optimizer ever switches to a hash-aggregate from a sort-aggregate).

junsklee:
Please add scenarios testing aggregates other than COUNT, like SUM and AVG.

kimhak-tech:
Updated:
- add `drop table if exists`
- remove `order by` clause
- add different aggregate functions (combination with different functions)
- add cci answer (answer file is now different on cci)

junsklee:
For the AVG aggregate, you may also use https://www.cubrid.org/manual/internal/en/develop/sql/function/numeric_fn.html#ROUND or https://www.cubrid.org/manual/internal/en/develop/sql/function/numeric_fn.html#TRUNCATE if you wish to avoid varying answers in '.answer' (sql) and '.answer_cci' (sql_by_cci).

junsklee (approved):
LGTM. Well done 👍

junsklee (commented):
While not strictly required, capitalizing the first letters of the "evaluate" statements in tests 5–12 is likely good practice to ensure consistency with the rest of the tests.

kwonhoil (approved):

ssihil (approved):

tw-kang (approved):

PR #2271, opened 2025-07-29
[CBRD-26058] Add sql test case for Fix error in CASE expression with logical type

oychanphallin:
Refer to : http://jira.cubrid.org/browse/CBRD-26058

Issue:
Fix error in CASE expression with logical type.
A system error occurs when a `CASE` expression has one comparison branch and the other is `NULL` or returns no result. The system treats it as a logical type but fails to handle it properly. The fix ensures consistent casting to `INTEGER` to avoid the error.

**Test Coverage**
Following test scenarios are being cover:

1. Mixed CASE branches (Integer in THEN, NULL in ELSE)
2. THEN returns 0 or 1
3. THEN returns TRUE or FALSE
4. THEN returns string literals
5. THEN returns 1, 2, or NULL
6. THEN contains scalar subquery
7. CASE in SELECT list (comparison vs NULL)
8. THEN comparison predicate, ELSE NULL
9. THEN NULL, ELSE comparison predicate
10. CASE in ORDER BY (comparison vs NULL)


junsklee:
Please turn this to draft before I approve it

junsklee:
The scenario description comments should be an evaluate statement. Also, there is no need to add in the "Scenario" portion, as it adds unnecessary clutter.

E.g., 
as-is:
```
-- =====================================================
-- Scenario 1 - Mixed CASE branches (Integer in THEN, NULL in ELSE)
-- =====================================================
```
to-be:
```
evalulate '1 - Mixed CASE branches (Integer in THEN, NULL in ELSE)';
```
``

junsklee:
Your test file in the test folder should have the '.sql' extension, and not the '.answer' extension. Right now, you have two answer files.

junsklee:
For your test and answer files, you should use underscores between 'cbrd' and '26058'.

as-is:
`sql/_13_issues/_25_2h/cases/cbrd-26058.answer`
to-be:
`sql/_13_issues/_25_2h/cases/cbrd_26058.sql`

junsklee:
You can use CASE statements in other parts of the query, not just in the where statement.

For example,
```
SELECT a,
       CASE WHEN a=1 THEN 'one'
            WHEN a=2 THEN 'two'
            ELSE 'other'
       END
FROM case_tbl;
```

oychanphallin:
> ```
> evalulate '1 - Mixed CASE branches (Integer in THEN, NULL in ELSE)';
> ```

@junsklee Unfortunately, this results in an error because the `evaluate` statement is not supported.
Can i use this instead?
```
-- evalulate 1 - Mixed CASE branches (Integer in THEN, NULL in ELSE)
```

oychanphallin:
Updates include:
- Updated naming convention
- Updated comment sections
- Added `evaluate 7`

junsklee:
> > ```
> > evalulate '1 - Mixed CASE branches (Integer in THEN, NULL in ELSE)';
> > ```
> 
> @junsklee Unfortunately, this results in an error because the `evaluate` statement is not supported. Can i use this instead?
> 
> ```
> -- evalulate 1 - Mixed CASE branches (Integer in THEN, NULL in ELSE)
> ```


Looks like there was an extra 'l' in evaluate. Here is the evaluate statement used in csql:
```
csql> evaluate '1 - Mixed CASE branches (Integer in THEN, NULL in ELSE)';

=== <Result of EVALUATE Command in Line 1> ===

  Result              
======================
  '1 - Mixed CASE branches (Integer in THEN, NULL in ELSE)'

Execute OK. (0.020001 sec) Committed. (0.000000 sec) 

1 command(s) successfully processed.

```

sql/_13_issues/_25_2h/cases/cbrd_26058.sql
```diff
@@ -0,0 +1,99 @@
...
+SELECT DISTINCT dummy
+FROM dual T1
+WHERE
+    CASE
+        WHEN T1.dummy IN ('X', 'Y', 'Z')
+        THEN (SELECT 1) = 1
+        ELSE FALSE
+    END = TRUE;
+
+-- evalulate 7 - CASE expression in SELECT clause
```
junsklee: Not sure if you've grasped the JIRA issue fully.. This test case has little to do with the issue statement "A system error occurs when a CASE expression has one comparison branch and the other is NULL or returns no result."

junsklee:
```suggestion
evaluate '7 - THEN comparison predicate, ELSE NULL';
-- THEN : comparison predicate (dummy != 'X')
-- ELSE : NULL
SELECT DISTINCT dummy
FROM dual
WHERE
    CASE
        WHEN dummy = 'X' THEN dummy != 'X'
        ELSE NULL
    END = TRUE;
```

junsklee:
```suggestion
evaluate '7 - CASE in SELECT list (comparison vs NULL)';
-- THEN : comparison predicate (dummy = 'X')
-- ELSE : NULL
SELECT dummy,
       CASE
           WHEN dummy = 'X' THEN dummy = 'X'
           ELSE NULL
       END AS flag
FROM dual;
```

junsklee:
I'd suggest adding in a test case that has a comparison predicate in the _THEN_ statement, and _NULL_ as the ELSE statement.
```
evaluate '8 - THEN comparison predicate, ELSE NULL';
-- THEN : comparison predicate (dummy != 'X')
-- ELSE : NULL
SELECT DISTINCT dummy
FROM dual
WHERE
    CASE
        WHEN dummy = 'X' THEN dummy != 'X'
        ELSE NULL
    END = TRUE;
```

junsklee:
I also suggest adding in a TC that has NULL as the THEN statement, and ELSE has a comparison predicate:
```
evaluate '9 - THEN NULL, ELSE comparison predicate';
-- THEN : NULL
-- ELSE : comparison predicate (dummy = 'X')
SELECT DISTINCT dummy
FROM dual
WHERE
    CASE
        WHEN dummy != 'X' THEN NULL
        ELSE dummy = 'X'
    END = TRUE;
```

junsklee:
Include a test case with a CASE statement in the ORDER BY section.

```
evaluate '10 - CASE in ORDER BY (comparison vs NULL)';
-- THEN : comparison predicate (dummy = 'X')
-- ELSE : NULL
SELECT dummy
FROM dual
ORDER BY CASE
             WHEN dummy = 'X' THEN dummy = 'X'
             ELSE NULL
         END;
```

junsklee:
Please add your PR description comment at the top of the sql file. Refer to previously added PRs to see the general formatting used and the content included (e.g., issue number, description, tests covered, etc.).

oychanphallin:
Updates include :
- Expanded test coverage (evaluate 7, 8, 9, and 10)
- Added a test description at the top of the SQL file
- Corrected the misspelled keyword `evaluate`

kwonhoil (approved):

junsklee (approved):

ssihil (approved):

bagus-kim (approved):

tw-kang (approved):

oychanphallin:
Hi @HyunukLee, Could you please take a quick look and review this? thank you.

HyunukLee (approved):

PR #2285, opened 2025-08-05
[CBRD-25709] Maximum trigger depth does not work

LyhengCHHAY:
Refer: http://jira.cubrid.org/browse/CBRD-25709

Issue: `Infinite loop occurs when a recursive structure occurs in the Trigger DEFERRED statement (Maximum trigger depth does not work)`
Tested Ver. `CUBRID 11.5 (11.5.0.1893-61bef33) (64bit release build for Linux)`

Added a parameter to check the number of depth of deferred trigger


sql/_13_issues/_25_2h/cases/cbrd_25709.sql
```diff
@@ -0,0 +1,16 @@
...
+
+-- Trigger with depth check
+CREATE TRIGGER trg_a
+DEFERRED INSERT ON tbl_a
+IF NEW.depth < 32
+EXECUTE INSERT INTO tbl_a(col_a, col_b, depth) VALUES (1, 2, NEW.depth + 1);
+
+-- Trigger the recursion
+INSERT INTO tbl_a(col_a, col_b) VALUES (1, 1);
+
```
junsklee, line 16: Please insert your cleanup (DROP) statements

junsklee:
Let me know if this is still in progress; if not, the test case could benefit by implementing more scenarios. There seems to be no SELECT statement either, so I'd check that out as well.

PR #2420, opened 2025-11-27
[CBRD-26245] Create new TC to fix filteredkeys always reported as 0 in TRACE statistics

LyhengCHHAY:
Refer : http://jira.cubrid.org/browse/CBRD-26245

### DIFF
--- a/sql/_13_issues/_25_2h/cases/cbrd_26245.sql
+++ b/sql/_13_issues/_25_2h/cases/cbrd_26245.sql
@@ -1,16 +1,16 @@
 set trace on;
 
-DROP TABLE IF EXISTS tbl1;
+DROP TABLE IF EXISTS tbl_w;
 
-CREATE TABLE tbl1 (
+CREATE TABLE tbl_w (
     a INT,
     b CHAR(3),
     c INT
 );
 
-CREATE INDEX idx_tbl1_ab ON tbl1 (a, b);
+CREATE INDEX idx_tbl_w_ab ON tbl_w (a, b);
 
-INSERT INTO tbl1 VALUES
+INSERT INTO tbl_w VALUES
     (2, 'ABC', 15000),
     (2, 'ZZZ', 13000),
     (3, 'JJJ', 17000),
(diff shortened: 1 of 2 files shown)


sql/_13_issues/_25_2h/cases/cbrd_26245.sql
```diff
@@ -0,0 +1,111 @@
...
+
+INSERT INTO tbl_w VALUES
+    (2, 'ABC', 15000),
+    (2, 'ZZZ', 13000),
+    (3, 'JJJ', 17000),
+    (3, 'KKK', 14000),
+    (4, 'ABC', 18000),
+    (4, 'III', 13000);
+
+evaluate '[select * from tbl_w where a > 1 and a < 5 and b < 'K' and c > 10000]';
```
junsklee, line 21: This evaluate statement is incorrectly returning an error, likely due to the use of single quotes in `'K'`, and needs to be corrected.

sql/_13_issues/_25_2h/cases/cbrd_26245.sql
```diff
@@ -0,0 +1,111 @@
...
+CREATE INDEX idx_tbl_x_ab ON tbl_x (a, b);
+
+INSERT INTO tbl_x VALUES
+    (1, 'AAA',  5000),
+    (2, 'BBB',  6000),
+    (3, 'CCC',  7000),
+    (4, 'DDD',  8000),
+    (5, 'EEE',  9000);
+
+evaluate '[select * from tbl_x where a > 1 and a < 6 and b = 'CCC' and c >= 7000]';
```
junsklee, line 48: This evaluate statement is incorrectly returning an error, likely due to the use of single quotes in `'CCC'`, and needs to be corrected.

sql/_13_issues/_25_2h/cases/cbrd_26245.sql
```diff
@@ -0,0 +1,111 @@
...
+
+INSERT INTO tbl_z VALUES
+    (2, 'AAA', 11000),
+    (2, 'BBB',  9000),
+    (2, 'CCC', 15000),
+    (3, 'DDD', 13000),
+    (3, 'EEE',  8000),
+    (3, 'FFF', 16000);
+
+evaluate '[select * from tbl_z where a = 2 and b <= 'CCC' and c > 10000]';
```
junsklee, line 98: This evaluate statement is incorrectly returning an error, likely due to the use of single quotes in `'CCC'`, and needs to be corrected.

junsklee:
The goal of this issue (CBRD-26245) is to verify that the `filteredkeys` value displays the proper number of filtered keys in queries' trace statistics (and that the `readkeys` value does not include the `filteredkeys` value). It seems that we cannot verify the values of filteredkeys using CTP's sql function because query plans' results convert numeric values into question marks (?). I suggest that this issue be recreated as a shell test case, and verify just the two aforementioned values (and other values should be ignored as they may render the TC unstable).

PR #2520, opened 2026-03-17
[CBRD-26504] Add SQL test case for incorrect propagation of START WITH as WHERE clause

kimhak-tech:
Refer to http://jira.cubrid.org/browse/CBRD-26504

**Issue**: When `single_table_opt` is active, `qo_reduce_equality_terms` incorrectly propagates `START WITH` conditions as `WHERE` conditions, causing hierarchical queries to return wrong results.


ssihil (approved):

kwonhoil (approved):

junsklee (approved):

sql/_13_issues/_26_1h/cases/cbrd_26504.sql
```diff
@@ -0,0 +1,112 @@
...
+
+DELETE FROM tbl WHERE c1 IN (
+    SELECT c1 FROM tbl
+    START WITH c1 = 'B'
+    CONNECT BY PRIOR c1 = c2
+);
+
+SELECT c1 FROM tbl ORDER BY c1;
+-- Expected: A, C (B and D deleted, A and C untouched)
+
```
bagus-kim, line 126: -- 이번 이슈가 reduce_equality_terms 와 직접 연결되어 있으므로,동등 조건(equality condition, = 형태 조건) 이 실제로 같이 존재하는 케이스가 있으면 좋습니다.Case6에서는 필터링이 됨으로 항상 TRUE인 케이스가 추가를 검토부탁드립니다

evaluate 'Case 11. START WITH and WHERE equality terms on same query';
SELECT c1 FROM tbl
WHERE c1 = c1
START WITH c1 = 'A'
CONNECT BY PRIOR c1 = c2
ORDER BY c1;
-- Expected: A, B, C, D

--c2컬럼으로 시작하는 경우도 고려

evaluate 'Case 12. START WITH on parent column';
SELECT c1 FROM tbl
START WITH c2 = 'A'
CONNECT BY PRIOR c1 = c2
ORDER BY c1;
-- Expected: B, C, D
-- START WITH on c2 must not be reduced or propagated like a WHERE equality term.

kimhak-tech: Reflected, thank you for the review.

bagus-kim (approved):

xmilex-git (approved):

PR #2988, opened 2026-07-01
[CBRD-25709] Add sql test case for Infinite loop occurs when recursion occurs in the Trigger DEFERRED statement

kimhak-tech:
Refer to http://jira.cubrid.org/browse/CBRD-25709

**CBRD-25709** fixes an issue in which recursive DEFERRED triggers bypassed the maximum trigger depth limit and caused an infinite loop instead of aborting the transaction.


sql/_13_issues/_26_2h/cases/cbrd_25709.sql
```diff
@@ -0,0 +1,171 @@
...
+ 
+evaluate 'Case 7. Bounded DEFERRED recursion whose own limit exceeds max depth (32)';
+CREATE TRIGGER tr_bounded_exceed
+DEFERRED INSERT ON t3
+IF ((SELECT COUNT(*) FROM t3) < 50)
+EXECUTE INSERT INTO t3 VALUES (1);
+ 
+INSERT INTO t3 VALUES (1);
+-- Expected: ERROR - Maximum trigger depth 32 exceeded
+
```
bagus-kim, line 138: 데이터를 확인하는 구문을 추가하는 것이 좋을 것 같습니다.

SELECT COUNT(*) FROM t3; --Expected: 0

sql/_13_issues/_26_2h/cases/cbrd_25709.sql
```diff
@@ -0,0 +1,171 @@
...
+DROP TABLE t3;
+ 
+evaluate 'Case 8. ROLLBACK before commit means DEFERRED action never fires';
+-- autocommit must be OFF here: otherwise the INSERT below would commit itself immediately
+autocommit off;
+ 
+CREATE TRIGGER tr1
+DEFERRED INSERT ON t1
+EXECUTE INSERT INTO t1 VALUES (1, 2);
+
```
bagus-kim: Case 8의 의도는 "ROLLBACK하면 DEFERRED 트리거가 실행되지 않는다"를 검증하는 것으로 이해했습니다.
그런데 현재 코드에서는 CREATE TRIGGER와 INSERT가 같은 트랜잭션 안에 있어서, ROLLBACK이 둘 다 취소합니다. 
COUNT(*)=0이 나올 때"DEFERRED 트리거가 실행되지 않아서"가 아니라, "트리거 자체가 존재하지 않게  되었기 때문"일 경우도 발생할 수 있을 것 같습니다.

그러므로, off를 trigger이후로 하는 것이 어떠할까요?
CREATE TRIGGER tr1
DEFERRED INSERT ON t1
EXECUTE INSERT INTO t1 VALUES (1, 2);

;autocommit off

INSERT INTO t1 VALUES (1, 1);
ROLLBACK;
SELECT COUNT(*) FROM t1;
--Expected: COUNT(*) = 0

;autocommit on

bagus-kim:
현TC는 Case 1(INSERT), Case 9(UPDATE)가 있고, DELETE까지 추가하면 좋을 것 같습니다.

Case 11: DEFERRED DELETE 트리거 재귀
evaluate 'Case 11. DEFERRED DELETE trigger recurses into itself';
INSERT INTO t1 VALUES (1, 1);

CREATE TRIGGER tr_del_self
DEFERRED DELETE ON t1
EXECUTE DELETE FROM t1 WHERE col1 = 1;

DELETE FROM t1 WHERE col1 = 1;
-- Expected A (0건 DELETE에도 트리거 실행): ERROR - Maximum trigger depth 32 exceeded
-- Expected B (0건 DELETE에는 트리거 미실행): 에러 없이 종료

DROP TRIGGER tr_del_self;
DELETE FROM t1;

--------
Case 6은 30회(깊이 제한 이하)에서 멈추고, Case 7은 50회(깊이 제한 초과)를 검증합니다. 정확히 32에서 멈추는 경계 케이스가 빠져 있어서, "32 초과"에서 에러가 나는지 "32 이상"에서 에러가 나는지 명확하지 않아 경계값 추가를 했으면 합니ㅏㄷ.


Case 12: 깊이 경계값(depth boundary) 검증 

evaluate 'Case 12. Bounded DEFERRED recursion stopping at exactly max depth boundary';
CREATE TABLE t3(col1 int);

CREATE TRIGGER tr_boundary
DEFERRED INSERT ON t3
IF ((SELECT COUNT(*) FROM t3) < 32)
EXECUTE INSERT INTO t3 VALUES (1);

INSERT INTO t3 VALUES (1);
SELECT COUNT(*) FROM t3;
-- "32 초과"에서 에러 발동 → 총 32행, 정상 종료
-- "32 이상"에서 에러 발동 → Error -528 발생

DROP TRIGGER tr_boundary;
DROP TABLE t3;

--------
Case 5(교차 테이블), Case 10(교차 이벤트 )은 검증하고 있지만, 같은 테이블의 같은 이벤트에 트리거가 2개  등록된 경우는 없는 것 같습니다.

Case 13: 같은 이벤트에 DEFERRED 트리거 2개 등록

evaluate 'Case 13. Two DEFERRED INSERT triggers on same table both causing recursion';
CREATE TRIGGER tr_a
DEFERRED INSERT ON t1
EXECUTE INSERT INTO t1 VALUES (1, 1);

CREATE TRIGGER tr_b
DEFERRED INSERT ON t1
EXECUTE INSERT INTO t1 VALUES (2, 2);

INSERT INTO t1 VALUES (0, 0);
-- Expected: ERROR - Maximum trigger depth 32 exceeded
-- INSERT → tr_a, tr_b 각각 실행 → 각 INSERT가 다시 tr_a, tr_b를 실행 → 증가 → 초과

DROP TRIGGER tr_a;
DROP TRIGGER tr_b;
DELETE FROM t1;

bagus-kim (approved):

sql/_13_issues/_26_2h/cases/cbrd_25709.sql
kimhak-tech: Thank you for the review, I've reflected your feedback in the test case.

That said, for Case 11 (DEFERRED DELETE self-recursion): when I ran it, it completed without error rather than hitting depth 32. This is because once the fixed target row (col1=1) is removed, there's nothing left to match, so the recursion naturally stops at depth 1. 

To address this, I updated the trigger to target a dynamically computed value via subquery (```WHERE col1 = (SELECT MAX(col1) FROM t1```)) instead, and pre-populated 32 rows to reliably trigger the depth-exceeded error. This now behaves as expected.

Also Case 10 includes a DEFERRED DELETE trigger paired with a DEFERRED INSERT trigger to cover the mixed insert/delete recursion scenario.

Thanks again for the thorough review.

bagus-kim: Thank you for taking the time to review my comments.

sql/_13_issues/_26_2h/answers/cbrd_25709.answer
```diff
@@ -0,0 +1,313 @@
...
+
+===================================================
+0
+
+===================================================
+0
+
+===================================================
+Error:-21019
+Cannot communicate with the broker or received invalid packet
```
ssihil: An unexpected error has occurred.
This appears to require verification.

sql/_13_issues/_26_2h/cases/cbrd_25709.sql
```diff
@@ -0,0 +1,229 @@
...
+-- (neither trigger fires itself directly, but together they form an
+--  infinite cycle through the deferred-activity queue; BFS-layer depth
+--  counting must catch cross-trigger cycles too, not just self-recursion)
+ 
+DROP TRIGGER tr1;
+DROP TRIGGER tr2;
+DELETE FROM t1;
+DELETE FROM t2;
+ 
+evaluate 'Case 6. Bounded DEFERRED recursion that stops before max depth (32)';
```
ssihil: Change Case 6 to 31, which is just before the max depth.
Change Case 7 to 33, which is just after the max depth.
Change Case 12 to 32, which is the max depth,
and how about moving Case 12 after Case 6 so we can check them all at once?

kangmin5505 (approved):

sql/_13_issues/_26_2h/answers/cbrd_25709.answer
kimhak-tech: I think Case 13 causes **exponential growth**, not linear growth. Both `tr_a` and `tr_b` are triggered by `DEFERRED INSERT ON t1`, and each trigger inserts into `t1` again. As a result, every insert recursively fires both triggers. That means depth 1 → 2 queued activities, depth 2 → 4, depth 3 → 8, and so on.

By the time it reaches depth 32, that's over a huge queued deferred activities, likely exhausting server memory/resources long before the depth-32 guard gets a chance to evaluate and raise an error.

What do you suggest in this case? Should I alter this test case?

ssihil: @kangmin5505 
위 현상 확인 부탁 드립니다.

kiho-um (approved):

sql/_13_issues/_26_2h/answers/cbrd_25709.answer
kangmin5505: @ssihil 
@kimhak-tech is  right. This is exponential growth. It exhausts memory (OOM) before the depth-32 guard can fire, so we get `-21019` instead of the depth error.
I think this is a separate issue from CBRD-25709 (whose scope was the infinite loop), so let's track the OOM case separately and alter this test to a single self-recursive trigger.

kangmin5505: This issue will be resolved in http://jira.cubrid.org/browse/CBRD-27032.

sql/_13_issues/_26_2h/cases/cbrd_25709.sql
```diff
@@ -191,30 +205,14 @@ DELETE FROM t1 WHERE col1 = (SELECT MAX(col1) FROM t1);
...
-evaluate 'Case 13. Two DEFERRED INSERT triggers on same table both causing recursion';
+evaluate 'Case 13. Two independent DEFERRED INSERT triggers on the same table/event';
 CREATE TRIGGER tr_a
 DEFERRED INSERT ON t1
 EXECUTE INSERT INTO t1 VALUES (1, 1);
 
 CREATE TRIGGER tr_b
 DEFERRED INSERT ON t1
-EXECUTE INSERT INTO t1 VALUES (2, 2);
+EXECUTE INSERT INTO t2 VALUES (2, 2);
```
ssihil, line 215: The problem is expected to be resolved. It seems best to wait.
https://github.com/CUBRID/cubrid-testcases/pull/2988#discussion_r3541090844

sql/_13_issues/_26_2h/answers/cbrd_25709.answer
kimhak-tech: @kangmin5505 thank you for your confirmation.

sql/_13_issues/_26_2h/cases/cbrd_25709.sql
kimhak-tech: Reflected.

sql/_13_issues/_26_2h/cases/cbrd_25709.sql
kimhak-tech: @ssihil  I've altered Case 13 to keep two independent DEFERRED triggers on the same event, but changed `tr_b`'s action to target `t2` instead of `t1`. Both triggers still fire on the original insert, but only `tr_a`'s chain recurses (linear growth), so this stays within CBRD-25709's scope and avoids the exponential OOM path now tracked separately in **CBRD-27032**.

ssihil (approved):

kwonhoil (approved):

junsklee (approved):

PR #2998, opened 2026-07-02
[CBRD-26632] Add sql test case for core dump after login() and rollback in one transaction

kimhak-tech:
Refer to http://jira.cubrid.org/browse/CBRD-26632

**CBRD-26632** fixes an issue where calling `login()` to a user created in an uncommitted transaction, then rolling back, could leave invalid session user state and cause a core dump on later statements.


sql/_13_issues/_26_2h/cases/cbrd_26632.sql
```diff
@@ -0,0 +1,219 @@
...
+CALL login('dba') ON CLASS db_user;
+DROP USER user9;
+COMMIT;
+
+
+evaluate 'Case 14. State Preservation: login() does not alter the session commit mode';
+CALL login('user2') ON CLASS db_user;
+INSERT INTO dba.tbl1 VALUES (300);
+SELECT COUNT(*) FROM dba.tbl1;
+-- Expected: COUNT(*) = 0 before the rollback check
```
bagus-kim: Case 14에서 login()이 세션의 commit mode를 변경하지 않음을 검증하려는 의도로 이해했습니다. 테스트 로직과 answer 파일은 정확합니다. 
그런데 SQL 파일의 주석이 answer 파일과 일치하지 않습니다. 첫 번째 SELECT의 주석은 "COUNT() = 0"이라고 되어 있지만, answer 파일에는 COUNT()=1로 기록되어 있습니다.
하단과 같이 수정을 하는 것이 어떠할까요?

-- Expected: COUNT(*) = 1 (INSERT is visible within the same uncommitted transaction)

sql/_13_issues/_26_2h/cases/cbrd_26632.sql
```diff
@@ -0,0 +1,219 @@
...
+ROLLBACK;
+SELECT COUNT(*) FROM dba.tbl1;
+-- Expected: COUNT(*) = 0 same before rollback
```
bagus-kim:
```suggestion
-- Expected: COUNT(*) = 0 (INSERT was rolled back, proving autocommit is still OFF)
```

bagus-kim (approved):

sql/_13_issues/_26_2h/cases/cbrd_26632.sql
kimhak-tech: Reflected. Thank you for spotting it out.

kwonhoil (approved):

junsklee (approved):
Mostly LGTM. Below are non-blocking suggestions.

**Additional Coverage Suggestions (All Non-Blocking)**

* `cases/cbrd_26632.sql:10, :46` — Coverage item 2 in the header states that the test **"does not change the session's current user."** However, Case 2 only verifies whether `CREATE USER` and `DROP USER` succeed, which serves only as indirect evidence of the current user's privileges. There is no statement anywhere in the file that directly checks the current session user.

  Considering that this core dump occurred because the user name became `NULL` at `authenticate_context.cpp:703`, adding `SELECT CURRENT_USER;` immediately after the failed `LOGIN` attempt around line 35, as well as at the login transition points in Cases 12 and 13, would provide the most direct regression coverage.

  It is also recommended to revise the coverage description so that it accurately reflects what the test actually verifies.


* `cases/cbrd_26632.sql:208-209` — Case 14 is the only case that explicitly performs a `ROLLBACK` after a successful `LOGIN`, but the only subsequent verification is `SELECT COUNT(*)`.

  The core dump stack reported in JIRA followed the path:

  ```text
  SHOW TABLES
    → mq_virtual_queries()
    → authenticate_context.cpp
  ```

  Adding a statement that accesses the catalog or a system view, such as `SHOW TABLES;`, immediately after line 209 would directly exercise the same crash path under the **successful login + rollback** scenario and would make the coverage significantly stronger.

ssihil (approved):

kiho-um (approved):

ctshim (approved):

PR #3047, opened 2026-07-10
[CBRD-26309] Add test case for Estimate number of groups

kimhak-tech:
Refer to: http://jira.cubrid.org/browse/CBRD-26309

This test case verifies optimizer cardinality estimation for `GROUP BY` queries for **CBRD-26309**. The test checks that the optimizer estimates the number of groups using NDV statistics and selectivity, instead of treating grouped and ungrouped queries as having the same cardinality.

## Specification / Acceptance Criteria

- Single-table, single-column `GROUP BY` estimates the actual NDV.
- Multi-column `GROUP BY` may overestimate actual groups by multiplying NDVs, but the estimate must not exceed estimated row cardinality.
- Filtered `GROUP BY` estimates fewer groups than the equivalent unfiltered query by using the filtered-row estimate.
- `JOIN` `GROUP BY` applies the same group-count estimate after join and filter selectivity.
- Expressions, `DISTINCT`, `HAVING`, `LIMIT`, constants, `NULL`s, and boundary row counts keep the group estimate inside the row cardinality cap.

### DIFF
--- a/sql/_36_guava/cbrd_26309/answers/cbrd_26309.answer
+++ b/sql/_36_guava/cbrd_26309/answers/cbrd_26309.answer
@@ -1,0 +1,827 @@
+===================================================
+0
+===================================================
+0
+===================================================
+10000
+===================================================
+0
+===================================================
+0
+===================================================
+0
+===================================================
+0
+===================================================
+0
+===================================================
+0
+===================================================
+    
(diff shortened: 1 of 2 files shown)


kimhak-tech:
This PR has move to shell test case PR: https://github.com/CUBRID/cubrid-testcases-private-ex/pull/3649

PR #3148, opened 2026-07-29
[CBRD-26431] Add SQL test case for COUNT(alias) on subquery expression rewrite

kimhak-tech:
Refer to: [http://jira.cubrid.org/browse/CBRD-26431](http://jira.cubrid.org/browse/CBRD-26431)

  **Issue:**
  COUNT(alias) on a subquery arithmetic expression alias returns 0 instead of the actual count
  in CS-mode when the row count is >= 16696.

  **Expected:**
  COUNT(alias), COUNT(qualified_alias), and COUNT(*) all return the correct row count.


xmilex-git (commented):
This TC has no way to confirm that the parallel heap scan actually ran. CBRD-26431 only reproduces on the parallel path, so if the scan falls back to serial (worker reservation failure, or a change in the heap page count), every case still passes even on a pre-fix build.

Could we add `set trace on;` / `show trace;`, following the `cbrd_25447` precedent in the same `_36_guava` directory? CTP masks the numeric values with `?`, so the answer file stays stable, and the line below pins down both parallel entry and the gather mode (`BUILDVALUE_OPT`, which is where the CBRD-26431 fix lives). It also gives Case 2/3 a real control, since `parallel(0)` should make this line disappear.

```
(parallel workers: ?, heap time: ?..?, readrows: ?..?, rows: ?..?, gather: buildvalue)
```

Also, I don't think Case 4/5 are boundary tests. Parallel entry is decided by the number of heap data pages (`heap_get_num_data_pages`), not the number of result rows, so `where col1 <= 16695` leaves t1's page count unchanged and both cases still take the parallel path. A separate small table would be needed for a serial control.

kimhak-tech:
> This TC has no way to confirm that the parallel heap scan actually ran. CBRD-26431 only reproduces on the parallel path, so if the scan falls back to serial (worker reservation failure, or a change in the heap page count), every case still passes even on a pre-fix build.
> 
> Could we add `set trace on;` / `show trace;`, following the `cbrd_25447` precedent in the same `_36_guava` directory? CTP masks the numeric values with `?`, so the answer file stays stable, and the line below pins down both parallel entry and the gather mode (`BUILDVALUE_OPT`, which is where the CBRD-26431 fix lives). It also gives Case 2/3 a real control, since `parallel(0)` should make this line disappear.
> 
> ```
> (parallel workers: ?, heap time: ?..?, readrows: ?..?, rows: ?..?, gather: buildvalue)
> ```
> 
> Also, I don't think Case 4/5 are boundary tests. Parallel entry is decided by the number of heap data pages (`heap_get_num_data_pages`), not the number of result rows, so `where col1 <= 16695` leaves t1's page count unchanged and both cases still take the parallel path. A separate small table would be needed for a serial control.

Thank you for the review. Updated:
  - Added `set trace on; / show trace;` for all cases to confirm parallel path entry (gather: buildvalue). 
  - Removed Case 4/5 (boundary by row count filter) — as you noted, WHERE col1 <= 16695 doesn't change the heap page count, so both cases still take the parallel path.

xmilex-git (approved):

sql/_36_guava/cbrd_26431/cases/cbrd_26431.sql
```diff
@@ -0,0 +1,85 @@
...
+ *    8. NULL-producing expression alias - COUNT(expr) must skip NULLs
+ *    9. Simple column alias control - proves bug is expression-specific
+ */
+
+drop table if exists t1;
+create table t1 (col1 int);
+
+-- insert 16696 rows (the threshold where the bug manifests)
+insert into t1
+select rownum
```
bagus-kim: 16696행이 차지하는 페이지수가 대략 21~42개정도로 계산됩니다.
현 INSERT문장으로는 경계선 근처의 값임으로, 페이지 수가 threshold(32)를 확실히 초과하도록 보강하면 좋을 것 같습니다.(칼럼을 추가하여 행당 크기를 증가)

--예
create table t1 (col1 int, pad varchar(300));
...
insert into t1
select rownum, lpad('x', 300, 'x')
from table({0,1,2,3,4,5,6,7,8,9}) a,
     table({0,1,2,3,4,5,6,7,8,9}) b,
     table({0,1,2,3,4,5,6,7,8,9}) c,
     table({0,1,2,3,4,5,6,7,8,9}) d,
     table({0,1,2,3,4,5,6,7,8,9}) e,
     table({0,1,2,3,4,5,6,7,8,9}) f
limit 16696;

sql/_36_guava/cbrd_26431/cases/cbrd_26431.sql
```diff
@@ -0,0 +1,85 @@
+/**
+ *  This test case verifies CBRD-26431: COUNT(alias) on a subquery expression
+ *  is rewritten incorrectly, returning 0 instead of the actual count. The bug occurs in CS-mode.
+ *
+ *  Coverage:
+ *    1-3. Ticket repro: three hint combinations (RECOMPILE, RECOMPILE parallel(0),
```
bagus-kim: 1-3 주석은 "티켓 재현: 세 가지 힌트 조합 - 모두 올바른 카운트를 반환해야 함" 으로 해석됩니다.
Case2,3은 병렬 스캔OFF로 "병렬스캔을 끄면, 문제없이 동작한다"의 대조군 역할로 이해됩니다.

아래와 같이  1, 2-3으로 주석 분리하는 것이 어떠할까요?
*1. Ticket repro: RECOMPILE hint - parallel scan triggers, must return correct count
*2-3. Control: parallel(0) disables parallel scan - serial path must also return
       correct count (confirms bug is parallel-scan-specific)

bagus-kim (approved):

bagus-kim (approved):

sql/_36_guava/cbrd_26431/cases/cbrd_26431.sql
kimhak-tech: Reflected.

sql/_36_guava/cbrd_26431/cases/cbrd_26431.sql
kimhak-tech: Reflected.

sql/_36_guava/cbrd_26431/cases/cbrd_26431.sql
```diff
@@ -0,0 +1,87 @@
...
+ *    4. Different arithmetic expression (multiplication)
+ *    5. Mixed expression and plain column aliases
+ *    6. Other aggregates (SUM/AVG) on expression alias
+ *    7. Nested subquery with expression alias
+ *    8. NULL-producing expression alias - COUNT(expr) must skip NULLs
+ *    9. Simple column alias control - proves bug is expression-specific
+ */
+
+drop table if exists t1;
+create table t1 (col1 int, pad varchar(300));
```
kiho-um: 테이블명에 숫자가 들어가면 실행계획의 답지에 "?"로 표시가 됩니다.
숫자가 아닌 문자로 작성 해주세요.

ssihil:
I suggest additional scenarios.

evaluate 'Case 10: BIGINT expression alias under parallel scan';
select /*+ RECOMPILE */ count(x.s), count(s), count(*)
from (select col1 * cast(1000000000 as bigint) as s from t1) x;   -- expect 16696 16696 16696
show trace;

evaluate 'Case 11: NUMERIC(38,10) expression alias under parallel scan';
select /*+ RECOMPILE */ count(x.s), sum(x.s)
from (select cast(col1 as numeric(38,10)) / 3 as s from t1 where col1 <= 3) x;  -- expect 3 , 2.0000000000
show trace;

-- Ext-B: GROUP BY on the expression alias - a different parallel-aggregation
-- gather path than the scalar (gather: buildvalue) cases above.
evaluate 'Case 12: GROUP BY on expression alias, count(expr) per group (parallel)';
select /*+ RECOMPILE */ x.g, count(x.s)
from (select col1 % 4 as g, col1 + col1 as s from t1) x
group by x.g order by x.g;                                        -- expect 4 groups, each count 4174
show trace;

-- Ext-C: DISTINCT aggregation over an expression alias.
evaluate 'Case 13: count(distinct expression alias)';
select /*+ RECOMPILE */ count(distinct x.s)
from (select col1 % 100 as s from t1) x;                          -- expect 100
show trace;

-- Ext-D: MIN/MAX over an expression alias (completes COUNT/SUM/AVG coverage).
evaluate 'Case 14: MIN and MAX on expression alias under parallel scan';
select /*+ RECOMPILE */ min(x.s), max(x.s)
from (select col1 + col1 as s from t1) x;                         -- expect 2 , 33392
show trace;

ssihil (approved):

junsklee:
**NEEDS FIX** — the DISTINCT half of the engine fix (`px_heap_scan_result_handler.cpp:947-982`) is left with zero coverage, and the `Coverage:` header contradicts what the recorded answer shows for Case 2.

Status note: the only check reporting on this head commit is "Greptile Review" (success); the combined status is still `pending`, so there is no CI run evidence for `16f9fa27`. JIRA CBRD-26431 and the merged engine PR were both available for this review, and no bundle file was truncated.

Acknowledgements up front, so they are not re-litigated: xmilex-git's `set trace on;` / `show trace;` request and the removal of the old row-count "boundary" Cases 4/5 are correctly applied; bagus-kim's `pad varchar(300)` request is correctly applied and materially fixes the test. Without `pad`, 16,696 narrow rows occupy roughly 20-40 heap pages — straddling the parallel-entry floor; with `pad`, the heap is ~5.3 MB, i.e. a few hundred pages, so parallel entry is no longer marginal. The trace assertion then proves it actually happened.

## 1. Issue coverage

The merged fix https://github.com/CUBRID/cubrid/pull/6710 touches exactly one function, `result_handler<...>::write()`, in two independent branches:

- `px_heap_scan_result_handler.cpp:984-1008` — the non-DISTINCT branch, `COUNT(<expr>)`. This is the branch the ticket repro hits.
- `px_heap_scan_result_handler.cpp:947-982` — the `Q_DISTINCT` branch, `COUNT(DISTINCT <expr>)`. Same defect, same shape of fix (raw `operand->value.value.dbvalptr` replaced by `fetch_peek_dbval()` when the operand is not `TYPE_CONSTANT`).

Coverage of the first branch is good and precise:

- `cases/cbrd_26431.sql:35`, `:40`, `:50`, `:55`, `:65`, `:73` all rewrite to `count((x.col?+…))` style operands (`answers/cbrd_26431.answer:23`, `:45`, `:88`, `:110`, `:154`, `:176`) and record `16696`, where JIRA CBRD-26431 examples 1 and 2 show `0` on the pre-fix build. Fail-before / pass-after is established for Cases 1 and 2 directly from the ticket's own output, and for Cases 4/5/7/8 by identity of code path.
- `cases/cbrd_26431.sql:73` (Case 8) is the strongest single case: it exercises the *second* half of the non-DISTINCT hunk, the `DB_IS_NULL (db_value_p)` test at `px_heap_scan_result_handler.cpp:1000-1003`, which pre-fix evaluated a stale pointer. `16695 / 16695 / 16696` at `answers/cbrd_26431.answer:168` is the right assertion for it.
- `cases/cbrd_26431.sql:81` (Case 9) is a genuine negative control: it rewrites to `count(x.col?)` (`answers/cbrd_26431.answer:198`), i.e. the `TYPE_CONSTANT` fast path, and is expected to pass on both builds.

**Blocking gap — `sql/_36_guava/cbrd_26431/cases/cbrd_26431.sql:34-83`:** no scenario in this file, and none in the sibling `sql/_36_guava/cbrd_25447/cases/cbrd_25447.sql` (which has `select distinct`, but never `count(distinct …)`), reaches `px_heap_scan_result_handler.cpp:947-982`. Half of the shipped fix therefore has no regression barrier at all: reintroducing the raw `dbvalptr` read in the DISTINCT branch would leave every case in this PR green. The test must exercise the specific internal code path under test, and here the code path is documented by the engine patch itself.

This is not a new finding — ssihil already proposed it as Case 13. What I want to add is **why that one should be taken first**, out of the five proposed:

- **Case 13, `count(distinct x.s)` — take this one.** `PT_COUNT` with `option == Q_DISTINCT`, one aggregate and one outptr entry, satisfies every condition of the eligibility gate at `px_heap_scan_checker.cpp:38-101`, so by construction it lands in the untested branch. It also arrives with an independent oracle: over `col1` = 1..16696, `col1 % 100` has exactly 100 distinct values, computable without running anything.
- **Cases 10 (BIGINT) and 11 (NUMERIC(38,10))** re-run the *same* branch as Case 1 with a different `DB_VALUE` type. `fetch_peek_dbval()` → `DB_IS_NULL` → `curr_cnt++` is type-agnostic, so they add breadth, not a new branch.
- **Case 12 (GROUP BY on the expression alias)** cannot reach the fix: the gate requires `xasl->type == BUILDVALUE_PROC` (`px_heap_scan_checker.cpp:45-48`), and a GROUP BY query is not one. It would gather through the list path instead — compare `sql/_36_guava/cbrd_25447/answers/cbrd_25447.answer:323`, where the GROUP BY scenario records `gather: mergeable list`. Worth having as parallel-scan breadth, but it is not CBRD-26431 coverage.
- **Case 14 (MIN/MAX)** is unproven: the gate rejects any aggregate that is not `PT_COUNT_STAR`/`PT_COUNT` (`px_heap_scan_checker.cpp:86-89`), yet Case 6 uses SUM/AVG and still records `gather: buildvalue` at `answers/cbrd_26431.answer:137` — so the runtime behaviour is broader than the gate reads. Its value can only be settled by running it and reading the trace; Case 13's cannot fail to be relevant.

When Case 13 is added, keep `show trace;` on it and confirm the block records `gather: buildvalue`. If it does not, say so in a comment rather than leaving the case in as implied coverage. Optionally give it a NULL-producing variant, mirroring Case 8, to also cover the `DB_IS_NULL` test at `px_heap_scan_result_handler.cpp:964-967`.

## 2. Determinism

No blocking issue. Recorded for the author's benefit:

- **Row values are pinned, not merely counted.** `cases/cbrd_26431.sql:22-30` inserts `rownum` over a cross-join of literal `table({...})` collections, so `col1` holds exactly 1..16696 regardless of fetch order — and, unlike the sibling's `db_class` cross-join, it does not depend on catalog contents. Case 6's `where col1 <= 3` is therefore always rows 1,2,3, and Case 8 nulls exactly one row.
- **No `ORDER BY` is needed anywhere**; every asserted statement returns a single scalar-aggregate row.
- **`set trace on` / `set trace off` are paired and scoped** (`cases/cbrd_26431.sql:32`, `:85`).
- **The `(parallel workers: …)` line is the sole discriminator between Case 2 and Case 3.** CTP masks digits, so the `rewritten query:` lines at `answers/cbrd_26431.answer:45` and `:67` are byte-identical (`PARALLEL(?)` covers both `PARALLEL(0)` and any other value). The only textual difference between the two cases in the whole answer file is the presence of the parallel line at `:50` versus its absence at `:69-72`. Please do not let these trace blocks be trimmed in a later cleanup — without them the two cases become indistinguishable.
- **Residual fragility, non-blocking:** the parallel heap scan falls back to serial whenever the worker manager cannot reserve workers, which is a runtime, load-dependent condition. If that happens the parallel line disappears and the case fails. That is the safe failure direction — a loud FAIL rather than a silent PASS on an unpatched build — and it matches the accepted `cbrd_25447` precedent, so I am not asking for a change. It is worth knowing as the first hypothesis if this file ever goes intermittent in CI.

## 3. Expected-result accuracy

- The answer file is CTP-generated: `===` separators, statement-count blocks at `answers/cbrd_26431.answer:1-8` (drop 0 / create 0 / insert 16696 / `set trace on` 0) and `:206-209`, masked digits throughout. No hand-written smells. 
- Values check out independently: `16696` matches `limit 16696`; Case 8's `16695 / 16695 / 16696` at `:168` matches "exactly one row nulled"; Case 6's `6` and `2.0` at `:124` match sum/avg over rows 1,2,3.
- **No `.answer_cci` is required for this PR — I verified this rather than inferring it.** The sibling `sql/_36_guava/cbrd_25447` does ship one, but the two files differ on exactly five lines, all `median()` output (`2002.5` versus `2002.5000000000000000`); every trace block is byte-identical across drivers. Checking how the CCI driver renders doubles elsewhere in the repo (`sql/_04_operator_function/_06_aggragate_func/_001_avg/answers/1001.answer` and `sql/_19_apricot/_07_analytic_clause/answers/_08_avg.answer` against their `.answer_cci` siblings): values needing more than ~9 decimals diverge (`0.7142857142857143` → `0.714285714`), while whole-number doubles such as `3.0` and `2.0` are byte-identical. Case 6's `avg(x.s)` = `2.0` therefore renders identically under both drivers, and nothing else in this answer is fractional.
- Answer-fix vs bug-report taxonomy: this is a new file recording post-fix correct behaviour. No product defect is being baselined.

**Finding — `sql/_36_guava/cbrd_26431/cases/cbrd_26431.sql:7-8`:** the header says

> `2-3. Control: parallel(0) disables parallel scan - serial path must also return correct count (confirms bug is parallel-scan-specific)`

but the recorded answer shows Case 2 taking the **parallel** path: `answers/cbrd_26431.answer:50` carries `(parallel workers: ?, …, gather: buildvalue)`. Only Case 3, where the hint sits inside the subquery, is serial (`answers/cbrd_26431.answer:69-72`). The per-case captions are correct — `:39` says "(parallel)" and `:44` says "(serial path)" — so it is the summary header alone that is wrong, and it is wrong about the one structural fact this suite turns on: which case is the serial control. This also matches CBRD-26431 itself, where example 2 (`parallel(0)` on the outer query) still reproduced the wrong result and only example 3 did not. I realise this wording came out of the review thread and was applied in good faith; the answer file simply disagrees with it. Suggested rewording:

```
 *    2. Control: parallel(0) on the outer query does NOT disable the parallel
 *       scan (ticket example 2) - must still return the correct count
 *    3. Control: parallel(0) inside the subquery takes the serial path -
 *       confirms the bug is parallel-scan-specific
```


## 4. Additional coverage suggestions

- After Case 13, Cases 10/11 (BIGINT, NUMERIC) are reasonable breadth. **If Case 11 is adopted, check the CCI rendering before merging**: its expected `2.0000000000` is a fractional NUMERIC, which is precisely the value shape that forced `sql/_36_guava/cbrd_25447/answers/cbrd_25447.answer_cci` to exist. That case, unlike anything in the current file, may require a `.answer_cci`.
- Case 12 is worth adding for parallel-scan breadth, but please label it as such — its expected gather mode is the list path, not `buildvalue`, so a reader should not count it as CBRD-26431 coverage.
- Not requested, just noted: a same-shape scenario at a row count *below* the parallel-entry page floor would be a true serial boundary control without relying on a hint. That needs a second, small table, as xmilex-git already pointed out. Optional.

## 5. Suite-level design assessment

The suite has a genuine independent oracle, so it does not self-baseline — no suite-level blocking finding.

- In every COUNT case, `count(*)` sits in the same result row as the fixed `count(<expr>)` and is produced by the `PT_COUNT_STAR` branch at `px_heap_scan_result_handler.cpp:939-943`, which the patch did not touch. A wrong `count(expr)` cannot enter the baseline unnoticed, because the untouched column disagrees with it in the same line.
- Case 3 (serial) and Cases 1/2 (parallel) assert the identical value `16696`, giving the equivalence-under-suppressed-optimization check the doctrine asks for.
- Case 6 (`6`, `2.0`) and Case 8 (`16695` vs `16696`) are hand-computable from `limit 16696` alone, independent of any build.
- Asserted values are not degenerate: `16696`, `16695`, `6`, `2.0` vary across scenarios, and no two scenarios share a byte-identical data block.

## Required actions

1. Add ssihil's Case 13 (`count(distinct <expression alias>)`) so the `Q_DISTINCT` branch at `px_heap_scan_result_handler.cpp:947-982` is covered, and confirm from its `show trace;` block that it records `(parallel workers: …, gather: buildvalue)`. If the trace shows it does not reach that handler, leave a comment stating the gap instead of leaving the case in as implied coverage.
2. Correct the `Coverage:` header at `sql/_36_guava/cbrd_26431/cases/cbrd_26431.sql:7-8` so it stops describing Case 2 as the serial control; Case 3 is the only serial case, per `answers/cbrd_26431.answer:50` versus `:69-72`.
3. Regenerate `answers/cbrd_26431.answer` after both changes, and check whether the newly added case introduces any fractional output that renders differently under the CCI driver (none of the current values do).

## Verification

Proven statically: file pairing and directory placement against the `_36_guava` release convention; header/caption/answer consistency (and the one contradiction above); setup and teardown completeness for the shared database; row-value determinism of the setup; the answer file's provenance and the arithmetic of every asserted value; which of the two patched branches each case reaches, from the engine patch and the eligibility gate; and — by diffing existing driver-variant answer pairs in this repository — that no `.answer_cci` is needed as the PR currently stands.

Not proven statically, and still requiring a run on a test machine: that each case actually enters the parallel heap scan on your build (this depends on runtime worker reservation, not just on heap size); that the added DISTINCT case reaches the patched handler; and the fail-before / pass-after claim itself. Please run this case against a pre-fix and a post-fix build and confirm the NOK → OK transition, with the pre-fix run showing `0` for the `count(<expr>)` columns exactly as CBRD-26431 reports.

kiho-um (approved):

sql/_36_guava/cbrd_26431/cases/cbrd_26431.sql
kimhak-tech: Reflected. Renamed t1 → tbl (no digits) in both .sql and .answer files.

kimhak-tech:
Reflected. Added all 5 suggested cases (10-14). Case 13 (count distinct) confirmed reaching the Q_DISTINCT branch via trace (gather: buildvalue). Also added .answer_cci for CCI driver rendering differences.

kimhak-tech:
Reflected. All 3 required actions addressed:

1. **Case 13 added** — `count(distinct x.s)` covers the `Q_DISTINCT` branch. Trace confirms `(parallel workers: ..., gather: buildvalue)`.
2. **Coverage header corrected** — Case 2 now correctly described as parallel path; Case 3 as the serial control.
3. **Answer regenerated** via CTP run. Verified on both JDBC and CCI — added `.answer_cci` for the NUMERIC(38,10) rendering difference in Case 11.

junsklee (approved):

kwonhoil (approved):

PR #3226, opened 2026-08-10
[CBRD-26991] Add SQL testcase for add_member NULL target error handling

kimhak-tech:
<http://jira.cubrid.org/browse/CBRD-26991>

### Purpose

`call add_member()` 등의 메서드를 NULL 대상에 호출하면, 수정 전에는 에러 없이 `Execute OK`만 반환되고 결과 집합이 없었습니다.
수정 후에는 파서 단계(리터럴 NULL)와 런타임 단계(세션 파라미터 NULL)에서 각각 에러를 반환합니다.

### Implementation

8개 케이스로 `add_member`/`drop_member`/`set_password` 메서드의 NULL 대상 처리를 검증합니다.
리터럴 `NULL` 대상(파서 경로, Error:-493)과 `find_user`가 반환한 런타임 NULL 대상(런타임 경로, Error:-495)을 모두 테스트하며, 정상 동작(유효한 대상)과 기존 에러(존재하지 않는 사용자, Error:-165)도 대조합니다.

### Remarks

- 케이스 8개(`.sql` 헤더의 Coverage 참조), `sql/_13_issues/_26_2h/` 배치
- 검증: `11.5.0.2407-7d51f79`, 결정성 3회 전부 PASS, CCI 교차 일치, `.answer`는 CTP 출력에서 승격
- fail→pass **confirmed**: pre-fix `11.5.0.2389-d7ff1bd` FAIL → `11.5.0.2407-7d51f79` PASS

<details><summary>fail→pass 상세·귀속 한계</summary>

pre-fix 11.5.0.2389-d7ff1bd FAIL → fixed 11.5.0.2407-7d51f79 PASS (one run each)

</details>


kiho-um:
- 본문 Implementation, Remarks에는 "7개 케이스로 검증"이라고 적혀 있는데요. 실제 .sql의 Coverage 주석과 evaluate 라벨은 1~8, 총 8개입니다.
- 의도한 케이스 분류 기준(예: "NULL 처리 검증 6개 + 대조군 2개"처럼 세부 구분)을 명시해주면 혼돈을 줄일 수 있을 것 같습니다.

kiho-um (approved):

bagus-kim (approved):

ssihil:
I would like to propose adding a scenario.

call add_member('u') on (select null);
call add_member('u') on cast(null as object);

ssihil (approved):

junsklee (commented):
✅ PASS — The test case validates the issue correctly; no blocking problems found.

Note: every point below is a non-blocking suggestion, and none of them requires regenerating `.answer` (dropping `groups dba` leaves the output unchanged). Points already raised in this thread (greptile's Case 6 duplication note, kiho-um's "7 cases" mismatch in the PR body, ssihil's proposed scenarios) are not re-raised — the notes below build on them instead.

## 1. Issue coverage

The merged engine PR (https://github.com/CUBRID/cubrid/pull/7548) adds guards at two distinct points, and this test case exercises both of them separately and correctly.

- Parser path — `csql_grammar.y`, `opt_on_target: ON_ primary`, check `$2 == NULL || (PT_VALUE && PT_IS_NULL_NODE)`
  → `cases/cbrd_26991.sql:34` (Case 3), `:37` (Case 4), `:50` (Case 8), expected `answers/cbrd_26991.answer:32, :38, :66` = `Error:-493` (reported at parse stage = `Syntax`).
- Runtime path — `execute_statement.c`, `call_method()`, fires only when `target->node_type == PT_NAME && meta_class == PT_PARAMETER`
  → `cases/cbrd_26991.sql:41` (Case 5), `:44` (Case 6), `:47` (Case 7), expected `answers/cbrd_26991.answer:48, :54, :60` = `Error:-495` (execution stage = `Execute`).
  `nullgrp` is created by `call find_user(...) to nullgrp` at `:40`, so it really is a session parameter and satisfies the `PT_PARAMETER` condition. The "(parser path)/(runtime path)" wording in the captions matches the path actually taken.
- Control cases — `cases/cbrd_26991.sql:28` (valid target → `answer:19` `null`) and `:31` (nonexistent user → `answer:26` `Error:-165` = `User "%s" is invalid.`). Both match the CSQL output in the JIRA description/comments and lock in that the new guards did not break the normal path.

Pre-fix/post-fix discrimination holds per case: before the fix, all six statements in Cases 3–8 returned `Execute OK` with no error, so those six cases must diff on an unpatched build. Cases 1–2 are identical before and after, so they carry no discrimination, but they are valid regression guards (no request to remove them).

### On ssihil's proposed scenarios (`on (select null)`, `on cast(null as object)`)

I agree with the direction, but **these should not be added and promoted into `.answer` as-is.** Neither form is caught by either merged guard.

- `cast(null as object)` is a `PT_EXPR` (CAST) node at reduce time, and constant folding happens after parsing, so it passes the parser guard, which only tests `PT_VALUE`. The runtime guard only looks at `PT_NAME` + `PT_PARAMETER`, so it passes there too.
- `(select null)` is a `PT_SELECT` node and likewise slips past both guards.

So on the current develop build these two statements most likely still behave exactly like the CBRD-26991 symptom (success with no error). Recording that output in `.answer` would freeze a product defect as the expected result (the bug-report category, not the answer-fix category), and the test case would then fail later, when the engine closes this hole.

Suggested sequence:
1. Run both statements on a build and confirm the actual behavior/output.
2. If they succeed with no error, confirm with the developer whether expression-form NULL targets are intentionally out of scope, and file a follow-up CBRD issue.
3. Then either (a) add the cases after the follow-up fix lands, or (b) if they are added now, include a rationale comment in the `.sql` header such as `-- TODO: CBRD-XXXXX — expression-form NULL targets are outside the current guards (parser checks PT_VALUE, runtime checks PT_PARAMETER only)`.

## 2. Determinism

- There is no multi-row SELECT, so no statement requires `ORDER BY`. The output consists only of the scalar `null`, the object display `_db_user`, `Error:-NNN`, and affected-row counts of `0`.
- `answers/cbrd_26991.answer:15` `_db_user` is a class-name display that carries no OID (volid|pageid|slotid) and no timestamp, and it matches the CSQL output in the JIRA comments (`Result / ====== / _db_user`). Nothing in `.answer` depends on catalog order or wall-clock time.
- The 23 statements in the file map 1:1, in order, to the 23 blocks in `.answer` (login → `0` → `0` → Case 1 label → ... → `0` → `0`). No missing label, no off-by-one.

## 3. Expected-result accuracy

- All three error codes agree with the message catalog and the JIRA repro output: `-165` (`User "%s" is invalid.`, JIRA `ERROR: User "x_usr" is invalid.`), `-493` (parse stage), `-495` (execution stage). `.answer` has the shape of a promoted CTP result (`===` separators, `Error:-NNN`) with no hand-written smells.
- (suggestion) `cases/cbrd_26991.sql:5-7` — the header describes the runtime path as "NULL bind-variable targets", but the guard actually fires for a session-parameter (`PT_PARAMETER`) name created by `call ... to <var>`, not for a bind/host variable. It would also help to state in one line why the same message ("Methods require an object as their target.") surfaces as `-493` in one path and `-495` in the other (rejected at grammar stage vs at execution stage) — otherwise someone may later conclude the differing codes mean the answer file is wrong and "fix" it.
- (please confirm) Removal of `.answer_cci` — the PR body says the CCI cross-check matched, but in the broker_tester (CCI) TO-BE output in the JIRA comments, even the **valid** target case `call add_member('member_user') on valid_user;` comes out as `FAIL(-495)`. That implies the `to <var>` session parameter does not survive into the next statement on the CCI/broker path, so the target evaluates to NULL. If the same thing happens in a `sql_by_cci` run, Case 1 (`answer:19` `null` → `-495`) and Case 2 (`answer:26` `-165` → `-495`) would diverge and the case would fail only under CCI. Please double-check that the comparison was made against output actually produced by a `sql_by_cci` run.
- (please confirm) This engine change turns statements that used to succeed silently into errors. If any existing test case calls a method on a NULL/uninitialized target, its `.answer` is now stale. It would be good to record in the PR the result of a grep over `sql/` for `on null` / `on class db_user to` patterns (state "none" if there are no hits).

## 4. Setup / cleanup

- Both created users (`test_user_26991`, `test_group_26991`) are dropped at `cases/cbrd_26991.sql:53-54`, and `.answer:68,70` record `0` for success. No table/view/serial/prepare/system-parameter change is made, so no schema state is left in the shared database. CTP keeps executing the remaining statements after an error, so the cleanup statements still run even if Cases 3–8 fail.
- `cases/cbrd_26991.sql:44` `set_password('newpw')` does not actually change any password because the target is NULL, and on an unpatched build it is a no-op per the JIRA output, so it does not contaminate account state in the shared database.
- (suggestion) `cases/cbrd_26991.sql:24` — the `groups dba` clause in `create user test_group_26991 groups dba;` is not a precondition for being an `add_member` target (any user can have members in CUBRID). Because of it, `test_user_26991`, added as a member in Case 1, indirectly inherits dba group privileges, so a failed cleanup would leave a DBA-privileged account in the shared database. `create user test_group_26991;` is safer and leaves the `.answer` output unchanged.
- (suggestion) `cases/cbrd_26991.sql:27, :40` — the session parameter names `grp` and `nullgrp` are generic. Every other identifier in this file carries the `_26991` suffix; using `grp_26991` / `nullgrp_26991` removes any chance of a name collision with other cases sharing the same connection. No effect on `.answer`.
- `cases/cbrd_26991.sql:20` — the leading `call login('dba','') on class db_user;` reads as ensuring the DBA privileges needed by `add_member`/`set_password`; since the file never switches users afterwards, the absence of a restore statement leaks nothing.

## 5. Additional coverage suggestions

- (suggestion) after `cases/cbrd_26991.sql:28` — Case 1 only verifies that `add_member` returned `null`; it never verifies that the membership was actually created. Given that the original title of CBRD-26991 is about the call producing no result, a scalar query that checks whether the method actually did its work would make the oracle independent of the displayed string. Example (exact syntax to be confirmed on a build):
  `select count(*) from db_user u, db_user g where u.name='TEST_USER_26991' and g.name='TEST_GROUP_26991' and g in u.groups;`
  It is a single scalar, so no `ORDER BY` is needed.
- (suggestion) placing the same query once more right after Case 7 (`cases/cbrd_26991.sql:47`, `drop_member` on a NULL target) would additionally lock in that the failed call did not touch the membership. A regression where a NULL-target call silently does partial work cannot be caught by the error code alone.

## 6. Suite-level design assessment

This is not a plan/trace suite, so no `.queryPlan` sidecar is needed, and the expected values are not self-baselined from the build under test. The expected error codes can be independently confirmed from evidence outside the build — the message catalog (`-165`/`-493`/`-495`) and the TO-BE output in JIRA — and by the "what would change if the fix were reverted" test, all six scenarios in Cases 3–8 flip from error to no-error. So there is no suite-level oracle weakness. Adopting suggestion 5 would add a second independent oracle (state change) on top of the error codes and make the design more robust.

## Verification still needed

- Confirmed statically: the 23:23 statement/answer mapping, the mapping between the two engine guards and the cases, the identity of the error codes, completeness of resource cleanup, and the absence of nondeterministic values.
- Not confirmed statically: the actual FAIL → PASS transition between the pre-fix and post-fix builds (the PR body states `11.5.0.2389-d7ff1bd` FAIL → `11.5.0.2407-7d51f79` PASS, but this is the author's own report and is not backed by a CI log), whether Cases 1–2 produce identical output under CCI, and the current real behavior of the two statements ssihil proposed.
- Please run this test case on a test machine against the pre-fix and post-fix builds and confirm the NOK → OK transition, and compare the `sql_by_cci` result in the same run to settle whether `.answer_cci` is really unnecessary.

junsklee (approved):

ctshim (approved):

kimhak-tech:
@ssihil 님이 제안하신 두 시나리오를 현재 post-fix 빌드(`11.5.0.2451-e6ed61e`)에서 실행해 보았습니다.

```sql
csql> call add_member('dba') on (select null);
Execute OK.
csql> call add_member('dba') on cast(null as object);
Execute OK.
```

두 구문 모두 에러 없이 `Execute OK`를 반환하며, 이는 CBRD-26991 수정 전의 증상과 동일합니다.

junsklee님의 분석에 따르면, 엔진 PR(CUBRID/cubrid#7548)의 가드가 이 두 형태를 잡지 못합니다:
- **파서 가드** (`csql_grammar.y`): `PT_VALUE` 노드만 검사 → `(select null)`은 `PT_SELECT`, `cast(null as object)`는 `PT_EXPR`(CAST)이므로 통과
- **런타임 가드** (`execute_statement.c`): `PT_NAME` + `PT_PARAMETER`만 검사 → 두 형태 모두 해당하지 않으므로 통과

현재 이 구문들의 출력(`Execute OK`)을 `.answer`에 기록하면, 향후 이 gap이 수정될 때 테스트가 깨지게 됩니다 (bug를 expected로 고정하는 결과).

따라서:
- 이 PR에는 추가하지 않겠습니다
- expression 형태의 NULL target에 대한 follow-up CBRD 이슈를 별도로 생성하겠습니다

좋은 시나리오 제안 감사합니다.

kimhak-tech:
junsklee님의 리뷰에서 요청하신 확인 사항에 대한 결과입니다.

### Verification results

| Check | Result |
|-------|--------|
| FAIL → PASS transition | **Confirmed**: `11.5.0.2389-d7ff1bd` NOK → `11.5.0.2407-7d51f79` OK |
| Cases 1–2 identical under CCI | **Confirmed**: `sql_by_cci` PASS on `11.5.0.2451-e6ed61e`, no `.answer_cci` needed |
| Stale existing tests | **None**: grep over `sql/` for `on null;` found no hits outside `cbrd_26991.sql` |
| ssihil's scenarios | **Confirmed**: both `on (select null)` and `on cast(null as object)` return `Execute OK` on post-fix build — they bypass both guards. Will not add to this PR; confirming with developer whether a follow-up CBRD is needed |

### Changes applied (commit bb7a444)

- Fixed header: "bind-variable" → "session-parameter", added -493 vs -495 explanation
- Removed unnecessary `groups dba` from `test_group_26991`
- Renamed `grp`/`nullgrp` → `grp_26991`/`nullgrp_26991`
- Added membership verification queries after Case 1 and Case 7
- Regenerated `.answer`

kwonhoil (approved):

